PR #1, opened 2019-12-04
Integrate bs-jest. Add tests for vector primitives.

parkerziegler:
This PR adds in [`bs-jest`](https://github.com/glennsl/bs-jest) for unit testing our vector primitives. I was deciding on whether or not to adopt `bs-jest` or [`rely`](https://reason-native.com/docs/rely/), but decided on the former for a few reasons:

1. We're going to be using the primitives on the JS side, or one step removed from the JS side, so testing their implementation using a native Reason testing framework didn't seem like the right fit for where this code actually gets run.
2. It would add a lot of native setup overhead with `esy`.
3. It seems like it uses Js_of_ocaml (JSOO) rather than BuckleScript to run tests on the JS side, and I don't really know too much about that 😬 

Overall I feel pretty good about these tests! There are a few things that bother me, namely:

1. `bs-jest` doesn't allow you to have more than one assertion per test. You can get around this by dumping your values into a tuple, but this doesn't work with some of the cases where I need to assert on multiple values with assertions like `toBeCloseTo`. For example:

```reason
Expect.(
   expect((Vector.div(~v1, ~s)) |> toBeCloseTo(/* This has to be a float, it can't be a tuple. However, Vector.div returns a tuple :( */))
```

I get around this by having a separate test for the x value and y value of a vector, but I'd love to hear any other approaches to making this more concise.
2. We still have to deal with the intricacies of JS floating point math. In general this will be fine, as small imprecision won't meaningfully affect animations (at least, I don't think), but that's the major reason why we need `toBeCloseTo` and `toBeSoCloseTo`.

cc/ @JoviDeCroock, thought I'd include you hear just in case you have any thoughts on things, but also feel free to ignore!


src/core/Math.gen.tsx
```diff
@@ -12,6 +12,11 @@ export const sq: (_1:number) => number = MathBS.sq;
 
 export const sqf: (_1:number) => number = MathBS.sqf;
 
+export const pow: (_1:{ readonly base: number; readonly exp: number }) => number = function (Arg1: any) {
```
parkerziegler: You can ignore, the `.bs.js` and `.gen.tsx` files in these diffs since they're autogenerated by BuckleScript / `genType`.

src/core/Vector.re
```diff
@@ -1,5 +1,6 @@
 // The core vector type.
 type vector('a) = ('a, 'a);
+type t('a) = vector('a);
```
parkerziegler: `type t` is a special concept in Reason / OCaml. Typically it is used to refer to the type of a module itself. In this case, the `type t` of the `Vector` module is the `vector` type itself.

bsconfig.json
```diff
@@ -1,16 +1,24 @@
...
+    },
+    {
+      "dir": "__tests__",
+      "subdirs": true,
+      "type": "dev"
+    }
+  ],
   "package-specs": {
-    "module": "es6",
+    "module": "es6-global",
```
parkerziegler: https://bucklescript.github.io/docs/en/build-configuration#package-specs

It appears that `es6-global` will allow for relative imports of `node_modules`, which apparently (?) is better for being able to use ESM in the browser. I'm a bit foggy on this part of the BuckleScript docs (glad Patrick is working on them), so @kitten if you understand this better I'd love any insight you may have.

package.json
```diff
@@ -14,10 +15,26 @@
...
+  "jest": {
+    "testMatch": [
+      "<rootDir>/__tests__/**/*_test.bs.js"
+    ],
+    "transform": {
+      "^.+\\.js$": "babel-jest"
+    },
+    "transformIgnorePatterns": [
+      "node_modules/(?!(@glennsl/bs-jest|bs-platform)/)"
+    ]
```
parkerziegler: This is a negative lookahead such that `babel-jest` **will** transpile these `node_modules`. Since our `bsconfig` specifies `es6-global` as the build target, we need to further transpile all BuckleScript output (ours and that of 3rd party libs) down to ES5. This is why we have this here.

bsconfig.json
kitten: Afaik it’s only used in development for browsers since it results in import statements from `./node_modules/pkg/...` explicitly, so not so great for bundling in all cases.

So if you’re planning to have a build system past Bucklescript, I’d stick to the usual 🙂

__tests__/core/Vector_test.re
```diff
@@ -0,0 +1,221 @@
...
+
+    it(
+      "should expose a function for adding two vectors composed of float values (y value)",
+      () =>
+      Expect.(
+        expect(Vector.addf(~v1=v1Float, ~v2=v2Float) |> snd)
+        |> toBeCloseTo(4.2)
+      )
+    );
+  });
```
parkerziegler: Here's an example of what I was referencing in the PR description – `Vector.addf` returns a tuple of floats, `(float, float)`. However, we can't just use `toEqual((12.6, 4.2))` due to floating point precision issues. So I have two separate tests, one for the x value with `toBeCloseTo` and one for the y value, also with `toBeCloseTo`. Curious to hear about other approaches here, given that `bs-jest` only allows one assertion per test. Also curious on folks' thought re: `toBeCloseTo` or `toBeSoCloseTo`. I used both in these tests, opting for the latter in cases where significant figures seemed to be important, but this is pretty arbitrary.

bsconfig.json
parkerziegler: 🆒 Sounds good, reverting back to `es6` makes sense to me too based on the BS output.

PR #2, opened 2019-12-05
Add forces directory and Gravity module. Add initial force and forceV calculations.

parkerziegler:
This PR adds support for our first force – gravity 🎉. Modeling a force like gravity is fortunately not too difficult now that our vector primitives are in place. The equation for calculating the gravitational force between two objects is:

```
Fg = Gm1m2/r^2 * uv
```

In the equation, `G` is the universal gravitational constant, `m1` and `m2` represent the masses of the two attracting bodies, `r` represents the distance between them, and `uv` represents the unit vector describing the path from m1 to m2.

I decided to make the API for this module include both a `force` and `forceV` function, which will be alike for all future modules. The idea is that `force` gives us the *scalar* value of the force in Newtons, while `forceV` gives us the actual force *vector*.

I also added some more unit tests in this PR to check the gravity calculations and to ensure our `Math.pow` function handles negative exponents. cc/ @littlemilkstudio @mariano-formidable if you want to take a peek at things feel free. It's all written in Reason, so also feel free to skip for now!


__tests__/force/Gravity_test.re
```diff
@@ -0,0 +1,53 @@
...
+
+    it(
+      "should correctly derive the force of gravity at the Earth's surface", () => {
+      let massEarth = 5.97 *. Math.pow(~base=10, ~exp=24);
+      let massObject = 25.;
+      let rEarth = 6.371 *. Math.pow(~base=10, ~exp=6);
+      Expect.(
+        expect(
+          Gravity.force(~m1=massEarth, ~m2=massObject, ~r=rEarth)
+          |> Printf.printf("%0.*f\n", 2),
```
parkerziegler: A little trick from OCaml – this more or less mimics the behavior of JS `Number.toFixed()`, where you can round a floating point number to a specific precision. When originally calculating this out, I was getting a number like 2.45283472e24. TIL that `.toFixed()` won't actually truncate numbers w/ greater than 21 places of precision 😂 :mild-panic:

src/core/Math.re
```diff
@@ -5,9 +5,10 @@ let sqf = x => x *. x;
 // Exponents.
 let rec pow = (~base, ~exp) => {
   switch (exp) {
-  | 0 => 1
-  | 1 => base
-  | _ => base * pow(~base, ~exp=exp - 1)
+  | 0 => 1.
+  | 1 => Utils.foi(base)
+  | n when n < 0 => 1. /. Utils.foi(base) *. pow(~base, ~exp=exp + 1)
```
parkerziegler: Handle negative exponents!

src/forces/Gravity.re
```diff
@@ -0,0 +1,16 @@
+let g = 6.67428 *. Math.pow(~base=10, ~exp=-11);
+let force = (~m1, ~m2, ~r) => g *. m1 *. m2 /. Math.sqf(r);
+
+let forceV = (~m1, ~m2, ~v1, ~v2) => {
+  // Derive the vector pointing from m1 to m2.
+  let v = Vector.subf(~v1, ~v2);
```
mariano-formidable: Super nit, is `Vector.subf(~v1, ~v2)` equivalent to `v1 - v2`?  If so, the resultant vector will be a vector that points from v2 -> v1, but your comment says it's a vector pointing from m1 -> m2.  

This may be irrelevant, but if the formula requires a vector that points from m1 -> m2, then the vector subtraction should be `Vector.subf(~v2, ~v1)`.

And if the it doesn't matter for the formula, then at least the comment should say it's a vectoring "pointing from m2 to m1" :D 

Here's a pic below to illustrate what I mean:
image

src/core/Math.re
mariano-formidable: Noice!  But question, do you think you'll ever need fractional exponent support?  Does reason's type casting let you enforce that Math.pow only supports integer exponents?

src/forces/Gravity.re
parkerziegler: Sharp as always @mariano-formidable! Thanks for pointing that out. You're right that the comment should get switched around. It doesn't matter too much (yet) for our purposes that `subf` is implemented as `v1 - v2`, but it is good to be precise in the comments. Fixing now!

parkerziegler:
Just for visibility to everyone here, I altered `~m1` and `~m2` to `~attractorMass` and `~moverMass`, and `~v1` and `~v2` to `~attractor` and `~mover`. After talking with Mariano I think that nomenclature makes _way_ more sense and better conveys the directionality encoded in how `forceV` is implemented.

src/core/Math.re
parkerziegler: This more has to do with supporting operations like calculating the value of the universal gravitational constant `G`, which is 6.67 * 10^-11. AFAIK, 10^-11 is equivalent to 1/10^11, which is also equivalent to 1/10 * 10^-10. That's what this function is supposed to do. Also, the type of this function does indeed constrain exponents to being integers: https://github.com/FormidableLabs/renature/blob/master/src/core/Math.rei#L7 In this definition, `exp` is defined as _having_ to be an `int`.

mariano-formidable: Noice!

PR #3, opened 2019-12-06
Add initial implementation for rAF. Configure the project for use with Parcel. Add TypeScript setup.

parkerziegler:
Alright, this is kind of a big one, so feel free to tap out of code review at any point. The key highlights coming through in this one:

- Initial implementation of `rAF`. This is take pretty closely from @littlemilkstudio's blog post here: https://formidable.com/blog/2019/input-smoothing/ with a few minor tweaks.
- Creation of an `applyForce` function. This function takes in a bunch of arguments about a particular object's current `position`, `velocity`, and `acceleration`, and derives its next `position`, `velocity`, and `acceleration` based on those values.
- TS setup and Parcel integration. This was ridiculously easy. We'll eventually be using `rollup` as our build tool to ship the library, so Parcel is really just here to allow us to have an easy `example` project to dev against. You can start the example project by running `yarn start:example`.

@bmathews had some great commentary for me this morning // yesterday afternoon on how we handle accounting for the passage of time in these animations. Unlike a traditional tween, these animations don't necessarily have a set duration yet – they take the time they take based on the physics applied. Eventually this'll be connected to a tween of sorts between a `from` value and a `to` value, but that's down the road. The core pieces I'd love to get feedback on are:

- Does the `rAF` implementation make sense to people?
- How do we like the way the time step factor into our `listener` passed to `rAF`? (I'll explain more in code comments below).

Here's an example of the state updates coming through in the `console` as `rAF` runs: 
image

The values are _tiny_ b/c this is gravity operating on small mass objects, but still cool to see. One concern I have is that this seems to be running pretty consistently at 30fps this morning (yesterday afternoon I was seeing everything at 60). Not sure if it's just my browser dropping down to 30fps for some reason, but I'd love to see what other people experience.


example/index.ts
```diff
@@ -0,0 +1,57 @@
+import { rAF, forceV, applyForce, vector as Vector, Entity } from "../src";
```
parkerziegler: Although this is currently the `example` project, a lot of this code is what I picture getting moved into `useGravity`.

example/index.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    if (steps > 0) {
+      for (let i = 0; i < steps; i++) {
+        entity = applyForceForStep(config, mover, entity);
+      }
+    } else {
+      entity = applyForceForStep(config, mover, entity);
+    }
+
+    return entity;
```
parkerziegler: So this is our time step piece. @bmathews and I were talking about this yesterday. `react-spring` does this interesting thing where, for each frame, it determines the ms difference between the previous frame and the current one. Then, for each millisecond, it applies the physics calculations. For example, if 16ms elapsed b/t the previous and last frame, it would run the spring acceleration calculations 16 times. B/c we're working with two dimensional vectors I think we may be getting bogged down here – `react-spring` just uses one dimension: https://github.com/react-spring/react-spring/blob/master/src/animated/FrameLoop.ts#L67-L83

@littlemilkstudio what is your advice for this piece of our animation equation? Basically, as I understand it, we want to "continually apply the force" to match real world physics as closely as possible. I could see us either using a single dimension for `useGravity` and then using two dimensions for `useGravity2d` or reducing the steps by half (so, for example, we apply the force every 2ms rather than every 1ms).

example/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    velocity: mover.velocity,
+    position: mover.position
+  });
+};
+
+const run = (config: UseGravityArgs) => {
+  const { start, stop } = rAF();
+
+  start((timestamp, lastFrame, mover) => {
+    const diffTime = timestamp > lastFrame + 64 ? 0 : lastFrame;
```
parkerziegler: This is a neat little trick from `react-spring`. Basically, if we dropped four or more frames, we just fast forward to applying the latest force.

example/index.ts
```diff
@@ -0,0 +1,57 @@
...
+  setTimeout(() => {
+    stop();
+  }, 1000);
+};
+
+run({
+  moverMass: 20,
+  attractorMass: 10000,
+  attractor: [500, 500]
+});
```
parkerziegler: This sort of configuration will be a combination of user-provided (part of the API) or derived from user input (i.e. the location of `attractor` will actually be determined by placing `mover` at (0, 0) and placing `attractor` at (r, 0) for 1d gravity. For `useGravity2d` we'll ask them to provide a 2d location for their `attractor`.

example/index.ts
littlemilkstudio: yea this is dope

src/rAF/index.ts
```diff
@@ -0,0 +1,49 @@
...
+  };
+
+  const draw = (timestamp: DOMHighResTimeStamp) => {
+    state.mover = state.listener(timestamp, state.lastFrame, state.mover);
+    console.log(state.mover);
+    state.lastFrame = timestamp;
+    state.animationFrameId = requestAnimationFrame(timestamp => {
+      draw(timestamp);
+    });
+  };
```
littlemilkstudio: Won't the exiting condition be computed in the `state.listener` most of the time? probably some precision threshold for each physics step?

If so shouldn't it have some way of communicating whether or not to schedule the next frame?

Or do you have something else in mind with how you'll consume this that bypasses this problem?

example/index.ts
littlemilkstudio: do you want some listener to either be added to the config, or be a separate param that observes an update to the `entity` changes?

example/index.ts
littlemilkstudio: This was basically gonna be my approach for making the lerp frame rate independent. Cool to see this is the solution that react-spring came to.

littlemilkstudio: I'm not sure if I'm following your question 100%. If we could chat at some point so I could get a better understanding of your proposed approaches, that would be awesome.

example/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    acceleration: mover.acceleration,
+    velocity: mover.velocity,
+    position: mover.position
+  });
+};
+
+const run = (config: UseGravityArgs) => {
+  const { start, stop } = rAF();
+
+  start((timestamp, lastFrame, mover) => {
```
littlemilkstudio: I fell like `mover` could be part of the run clojure here instead of being so tied to `rAF`. 

Would love to get your thoughts on how you got on this route.

src/core/Utils.re
```diff
@@ -1,2 +1,28 @@
...
+    state.listener = listener;
+  };
+
+  let next = value => {
+    state.accumulator = reducer(~accumulator=state.accumulator, ~value);
+    state.listener(state.accumulator);
+  };
+
+  {start, next};
+};
```
parkerziegler: I'm not really sure if I'm going to use this at any point. Basically, this an accumulator for floating point numbers. I originally thought I might need it, then took more of the approach that `react-spring` does when tracking animation state – the `mover` object itself keeps track of its own `acceleration`, `velocity`, `position`, etc., which gets tracked in top level `rAF` state.

src/core/Vector.rei
```diff
@@ -1,5 +1,6 @@
-[@genType]
+[@genType "Vector"]
```
parkerziegler: This allows you to alias how you want the type exported. However, as you can see above, it messes up generics 😭 I'll see what `@genType` has to help with this.

src/forces/Force.re
```diff
@@ -0,0 +1,23 @@
...
+
+  // Add the velocity to the position.
+  let nextPosition = Vector.addf(~v1=position, ~v2=nextVelocity);
+
+  {
+    acceleration: nextAcceleration,
+    velocity: nextVelocity,
+    position: nextPosition,
+  };
+};
```
parkerziegler: The money maker 🤑 This is the function that determines, given an object with velocity, acceleration, and position, what will its _next_ velocity, acceleration, and position be? Thinking now that I should pass in the `entity` as a record directly to this function.

src/rAF/index.ts
parkerziegler: That's a really good point. I like the API you suggested in Slack, where `start` returns the `stop` function such that the `listener` has a way to cancel things. I suppose I haven't quite thought enough about the stopping aspect of things. Like, the gravitational force is theoretically going on for forever. I think `react-spring` has some `isOvershooting` flag to know when the spring has reach its end, so I could have something similar to determine a condition like "the mover has reached (or nearly reached) the attractor". In either case I think you're right that the `stop` will have to move, this just felt like a fine place to put it for the time being.

example/index.ts
parkerziegler: Hmmm, this is a good callout. I'm not totally sure yet. I think I like the idea of passing in the configuration for deriving the gravitational force alone and keeping the `listener` separate from the configuration. I do think eventually we'll want to get to a place where the user does have some access to the `position` of the entity on a frame by frame basis so they can apply styling (or this'll be somethinng the hook does for us). Like we discussed on Slack, having a little example working will help make this a bit clearer.

example/index.ts
parkerziegler: So I ended up getting here when trying to think of how the :heck: I preserve state across frames. I agree that `run` is a better place for this – but at the time of writing `rAF` I didn't yet have `run` implemented, so it ended up living in there. I agree that keeping `rAF` isolated to being a scheduler is the better approach, and as I begin to write the `gravity` function (basically what `run` is), that state will move out of `rAF`.

parkerziegler:
FYI I'm going to merge this as is for now so I can Keep Moving ™️, but the major changes in the next PR will be beginning to work out the API ergonomics for a `gravity` function that'll take the place of `run`. This will hopefully move a lot of state out of `rAF` and into specific force functions. The other big change will be exposing the `stop` function to the `listener`, such that we can set some sort of threshold for stopping the animation.

PR #5, opened 2019-12-09
Initial implementation for the gravity function.

parkerziegler:
Alright, this is an exciting one – we're finally starting to see things on a screen! The major goal of this PR is to start abstracting out the `rAF` loop into a `gravity` function. This function is what `useGravity` will call under the hood. It provides us with two main entry points for accessing animation state.

1. The `onFrame` callback. We pass this into `gravity` as part of the configuration currently. This callback gets executed once per frame and is handed the `position` vector of the `mover`. This works really nicely for `transform`s, since we basically have all the information we need to position the mover. This will also be the entry point for applying various interpolations to handle values like `opacity` or color properties like `background`. The API looks like this:

```typescript
gravity({
   config: {
      // Set up masses, initialVelocity, etc.
   },
   onFrame: (position: Vector<number>) => {
      // Do what we want with position 🎉 !
     element.style.transform = `translate(${position[0]}px, ${position[1]}px)}`;
   }
});
```

2. The `stop` function. `stop` gets returned by invoking `gravity` and allows you to imperatively stop the animation at any time. The API looks like this:

```typescript
const [stop] = gravity({
   config: {
      // Set up masses, initialVelocity, etc.
   },
   onFrame: (position: Vector<number>) => {
      // Do what we want with position 🎉 !
     element.style.transform = `translate(${position[0]}px, ${position[1]}px)}`;
   }
});

setTimeout(() => {
   // Arbitrarily stop the animation after 2 seconds.
   stop();
}, 2000);
```

The `gravity` function is also responsible for stopping the animation in response to certain physics conditions. The first condition involves the `mover` object reaching an arbitrary distance from the `attractor`. This is user configurable, but the gist is: **If the `mover` is x pixels from the `attractor`, stop the animation**. This is similar to what `react-spring` does – they use an `isOvershooting` flag to stop the animation if the spring has overshot its origin. The other case we need to deal with is: **what if a `mover` gets really far away from an `attractor`**? This is more problematic for a `transform` use case, and I'm inclined to punt this decision for now. Basically, if a `mover` gets really close to an `attractor`, the force acting on it will become so strong that it gets rocketed off into empty space. `constrain` is an attempt at reducing this effect, but it's not quite working the way I expect right now. I'm inclined to push figuring this out a bit farther down the road until we settle on a more concrete `useGravity` API.

Here's the look of the example in a more or less "standard" configuration (`mover` has no initial velocity and is purely attracted to the `attractor`).

191209_gravity

And here's the look of the example with a `mover` that has an initial velocity (a bit to the right and down).

191209_gravity_initial_velocity


example/index.ts
```diff
@@ -1,55 +1,37 @@
...
+document.body.appendChild(el);
+
+const positionAttractor = [500, 500] as vector<number>;
+const attractorEl = document.createElement("div");
+document.body.appendChild(attractorEl);
+
+el.style.height = "10px";
+el.style.width = "10px";
+el.style.borderRadius = "50%";
+el.style.background = "steelblue";
```
littlemilkstudio: great color 👏

example/index.ts
```diff
@@ -1,55 +1,37 @@
...
+
+attractorEl.style.height = "25px";
+attractorEl.style.width = "25px";
+attractorEl.style.borderRadius = "50%";
+attractorEl.style.background = "orange";
+attractorEl.style.left = `${positionAttractor[0]}px`;
+attractorEl.style.top = `${positionAttractor[1]}px`;
+attractorEl.style.position = "absolute";
+
+const [stop] = gravity({
```
parkerziegler: So here's the main idea with the API that `gravity` exposes. @littlemilkstudio and I also discussed the idea of returning a `controller` here as well, which would have methods for synchronously updating some internal animation state (i.e. `attractor` position). This would be useful for creating really interactive 2D animations that follow mouse move, or allow things like pausing and resuming animations arbitrarily, but that feels like a more advanced feature right now.

example/index.ts
```diff
@@ -1,55 +1,37 @@
...
+  config: {
+    moverMass: 2000,
+    attractorMass: 100000000,
+    attractor: positionAttractor,
+    initialVelocity: [0.01, 0.05],
```
parkerziegler: I added an `initialVelocity` parameter as well – just a fun way to set our `mover` on some course before the force of gravity sucks it inward. This is really fun to play around with.

src/animation/gravity.ts
```diff
@@ -0,0 +1,119 @@
...
+    moverMass: number;
+    attractorMass: number;
+    attractor: Vector<number>;
+    initialVelocity?: Vector<number>;
+    threshold?: {
+      min: number;
+      max: number;
+    };
+  };
+  onFrame: (position: Vector<number>) => void;
```
littlemilkstudio: I'm wondering if `onFrame` should have velocity in the params. I think this could be useful for the user in several cases, but off the top of my head:

- Would allow for them to switch between forces, by preserving velocity between the swap
- Might be nice for animating certain properties. I know squash n stretch (scale) can be a common velocity based property.

example/index.ts
```diff
@@ -1,55 +1,37 @@
...
-};
-
-run({
-  moverMass: 20,
-  attractorMass: 10000,
-  attractor: [500, 500]
+  },
+  onFrame: ([x, y]) => {
+    el.style.transform = `translate(${x}px, ${y}px)`;
+  }
```
parkerziegler: This is where I figure we can apply interpolations to convert `position` to normalized values. For example, applying a `lerp` here might be really useful to get 0 - 1 values for properties like `opacity`. We may need to pass back more state for that to happen, however. I'm still not sold on whether or not we leave this up entirely to the user like `react-pose` does with "passive" values: https://popmotion.io/pose/learn/passive/

src/animation/gravity.ts
```diff
@@ -0,0 +1,119 @@
...
+};
+
+/**
+ * The gravity function. This function tracks the internal state of the
+ * attractor and the mover and starts the frame loop to apply the gravitational
+ * force. Finally, it returns the stop function to imperatively stop the
+ * animation at any time. The animation will stop when the mover has reached the
+ * attractor as well.
+ */
+const DEFAULT_THRESHOLD = { min: 10, max: 1000 };
```
parkerziegler: Really just a placeholder for now, this is still quite unknown / TBD.

src/animation/gravity.ts
```diff
@@ -0,0 +1,119 @@
...
+    const { min, max } = params.config.threshold || DEFAULT_THRESHOLD;
+
+    const distance = magf(
+      subf({ v1: params.config.attractor, v2: state.mover.position })
+    );
+
+    if (distance <= min || distance >= max) {
+      // Mover has reached attractor or mover has exceeded the limits, stop animation.
+      stop();
+    } else {
```
parkerziegler: Here's the physics condition for stopping the animation. If the `mover` gets really close to the `attractor` or really far away (i.e. out of the bounds of the threshold) we stop the animation. I'm not sure I totally like this. I think it makes sense for getting arbitrarily _close_ to the object, but I'd love for the `mover` to basically always be able to come back to the `attractor` (i.e. never fly away).

src/animation/gravity.ts
littlemilkstudio: Just a thought here too, and this is super minor, but do you think the concept `frames` is an internal concept that doesn't need to be exposed in the outward facing api?

maybe having an `onPositionChange` & `onVelocityChange` or `onUpdate` might seem a little friendlier to the user.

Just a thought tho. I think you can make an argument for what you have here as well that it's more transparent as to what's going on under the hood

src/forces/Gravity.re
```diff
@@ -2,16 +2,24 @@ let g = 6.67428 *. Math.pow(~base=10, ~exp=-11);
...
 
   // Derive the magnitude of the above vector.
   let mag = Vector.magf(v);
 
+  // Constrain the gravitational force if threshold constraints were supplied.
+  let distance =
+    switch (threshold) {
+    | Some(th) => Math.constrain(~low=fst(th), ~high=snd(th), mag)
+    | None => mag
+    };
```
parkerziegler: My attempt to constrain the gravitational force. Basically, if the `distance` between `mover` and `attractor` exceeds the range, we use the clamped distance to calculate the force to apply. I'm missing something here I think, because this will constrain the force but not necessarily the position of the `mover`.

src/animation/gravity.ts
littlemilkstudio: Yea might be nice to have an option for no threshold

example/index.ts
parkerziegler: Stolen from all the `d3` tutorials I did back in the day 😂

littlemilkstudio: 🔥

src/animation/gravity.ts
parkerziegler: These are all good shouts! I could definitely see the API being:

```typescript
onUpdate: ({ position, velocity }) => {
   // Mutate your DOM how you like with either property.
}
```

It still feels a bit early to say though. The question I'm really wrestling with is figuring out the bounds of what we do vs. what the user does. I feel like Popmotion strikes a nice balance – sane defaults with options for configuration and customization. `react-spring` feels more opinionated, where it just handles it all for you and you pass these opaque props onto its opaque rendering target. Ultimately, `onUpdate` may be kept internal to `useGravity` if we force a rendering target on users (i.e. `animated.*` elements) or it may be opened up if we don't!

littlemilkstudio: yea real good point. I'm excited to see where you take it once you get more into react land

PR #6, opened 2019-12-11
Add support for a gravity1D and gravity2D function. Begin to integrate React in the example project.

parkerziegler:
This PR adds support for two separate gravity functions, `gravity1D` and `gravity2D`. Why the separation? Right now, it's because I find it a bit easier to reason about, and because it maps more cleanly to the types of animations that I think users will want.

For `from` / `to` style animations where users are animating a CSS property from one value to another, we use `gravity1D`. This representation basically treats the `mover` / `attractor` relationship as existing on a line. A `mover` is pulled towards an `attractor` according to the force of gravity along that line, and we `stop` the animation once it has overshot the `attractor`. Here's a simple drawing:

IMG_20191211_104013

For more complex, 2D transformations, we opt to use `gravity2D`. Currently, `gravity2D` has not stopping condition – it'll keep going until the user calls the `stop` function returned by the `gravity2D` function. What I like about this approach is that it allows us to achieve cool behavior like orbit! Moreover, I think 2D gravitational movement is something we don't necessarily expect to stop in our animations, so I sort of like not having a physics constraint to it.

Here's an example of what this looks like in our example project. which now uses React!

191211_animation_pane

The other key piece of this PR is realizing that some of our math was wrong. Rather than clearing the acceleration on each frame, I was actually _accumulating_ acceleration. This isn't correct – `acceleration` = F / m, not `acceleration` = `prevAcceleration` + F / m. This was partly responsible for some of the weirdness I was seeing with bodies flying _way_ too far off into space.

I'll leave some more code comments to help folks understand some more of the architectural decisions I made around things here.


src/animation/gravity.ts
```diff
@@ -56,21 +39,33 @@ const applyForceForStep = (
...
  */
-const DEFAULT_THRESHOLD = { min: 10, max: 1000 };
 
-export const gravity = (params: GravityParams): [() => void] => {
+interface Gravity1DParams {
+  config: {
+    moverMass: number;
+    attractorMass: number;
+    r: number;
+    initialVelocity?: number;
```
parkerziegler: Now that we're modeling things along a line, we can give a user more intuitive inputs. They don't need to know anything about the vector math – just give us the pair of masses, the distance between masses, and an initial velocity if you feel like it. This seems simpler to me and will probably be more familiar for most users seeking a `from` / `to` style animation.

src/animation/gravity.ts
```diff
@@ -123,3 +114,64 @@ export const gravity = (params: GravityParams): [() => void] => {
...
+    if (steps > 0) {
+      // Apply the gravitational force once for each step.
+      for (let i = 0; i < steps; i++) {
+        state.mover = applyForceForStep(state, params.config.threshold);
+      }
+    } else {
+      state.mover = applyForceForStep(state, params.config.threshold);
+    }
+
+    params.onUpdate({
```
parkerziegler: Notice we have no physics condition stopping the animation for `gravity2D`. Right now, I can't think of a logical reason to do this in library land. We hand back the `stop` function to the user so they can stop it according to their own userland logic, but for us to stop it preemptively doesn't strike me right now as useful. The one case I can see is if they supply `translate: transform(x, y)` and they want a specific `from` / `to` value, but I'm going to punt on that for now since I think we can add a flag – something like `infinite` – to help ease this.

src/core/Math.re
```diff
@@ -19,6 +19,23 @@ let lerp = (~acc, ~target, ~roundness) =>
...
+    ~low1=Utils.foi(low1),
+    ~high1=Utils.foi(high1),
+    ~low2=Utils.foi(low2),
+    ~high2=Utils.foi(high2),
+  )
+  |> Utils.iof;
+
+// Normalization.
+let norm = (~value, ~low, ~high) =>
+  remapf(~value, ~low1=low, ~high1=high, ~low2=0., ~high2=1.);
```
parkerziegler: Beginning work for interpolation. `remapf` will allow us to do numeric interpolation for a specific range, and `norm` is just a shortcut to do `interpolation` to a 0-1 range.

src/forces/Force.re
```diff
@@ -7,11 +7,7 @@ type entity = {
 
 let applyForce = (~force, ~entity) => {
   // Derive the acceleration created by the force and add it to the current acceleration.
-  let nextAcceleration =
-    Vector.addf(
-      ~v1=entity.acceleration,
-      ~v2=Vector.divf(~v=force, ~s=entity.mass),
-    );
+  let nextAcceleration = Vector.divf(~v=force, ~s=entity.mass);
```
parkerziegler: The keybug fix – we don't want to accumulate `acceleration`, we just want to derive it on every frame to calculate the `nextVelocity`. Theoretically we don't even need to store it in `state` anymore! Leaving it for now in case it ends up being useful in some way.

tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "baseUrl": ".",
     "esModuleInterop": true,
     "jsx": "react",
-    "lib": ["dom", "esnext"],
+    "lib": ["esnext", "dom"],
```
parkerziegler: TS was yelling about `Object.entries` not existing, and I think it had to do with `dom` being the first target for `lib`. We want to target `esnext` to get access to all thee latest features.

example/index.tsx
```diff
@@ -0,0 +1,114 @@
+import React from "react";
+import ReactDOM from "react-dom";
+import { gravity1D, gravity2D, vector as Vector, normNumber } from "../src";
+
+import "./index.css";
+
+const positionAttractor: Vector<number> = [100, 100];
```
littlemilkstudio: I think this will make the center of the attractor be the center of the circle instead of the top left corner
```
const positionAttractor: Vector<number> = [150, 150];
```

src/animation/gravity.ts
```diff
@@ -86,32 +81,28 @@ export const gravity = (params: GravityParams): [() => void] => {
...
 
     /**
-     * Conditions for stopping the animation. If the mover is within a specified
-     * threshold of the attractor, the gravitational force will become so strong
-     * that its acceleration will increase by orders of magnitude each frame. Similarly,
-     * a mover may still get catapulted by the gravitational
+     * Conditions for stopping the physics animation. For single animations with
+     * a discrete from / to pair, we want to stop the animation once the mover has
+     * reached the attractor. We can know its done by checking that the mover has
+     * not overshot the attractor.
```
mariano-formidable: Super nit, am I mis-reading the last sentence of this comment?  Is the "not" in "has not overshot" correct?  Isn't it supposed to be, it's done when it "has overshot the attractor"?

parkerziegler: Yep, just pushed a commit to fix that 😅

example/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+      attractorPosition: positionAttractor,
+      initialMoverVelocity: [0.1, 0],
+      threshold: {
+        min: 10,
+        max: 200
+      }
+    },
+    onUpdate: ({ position: [x, y] }) => {
+      if (m1.current) {
+        m1.current.style.transform = `translate(${x}px, ${y}px)`;
```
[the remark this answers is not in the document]
parkerziegler: Ah good catch!

example/index.tsx
parkerziegler: Yep yep, nice.

PR #7, opened 2019-12-12
Add support for basic number and color interpolators.

parkerziegler:
Feel free to skip review if you've already used your two hours on me this week // have more important things going on!

This PR looks quite a bit bigger than it is, so don't be too intimidated by the size! You can skip over all of the `.bs.js` and `.gen.tsx` files (and even the `.rei` files as well).

The core changes in this PR:

- Begin to restructure how we handle interpolating values. I'm moving most of these into `src/interpolate`. Currently this just includes `Remap_number`, `Lerp_number`, and `Lerp_color`, which handle operations like mapping an input domain to an output range, normalizing a range down to a [0, 1] domain, and linearly interpolating between a start and end color.
- Add a new example to show color interpolation (bottom left corner of the example page).
- Add an `index.ts` file to each folder. This helps keep our library organized such that `src/index.ts` can just be re-exporting all of these `index` files. In this way, a folder's particular `index` file becomes its public API. This will get restricted significantly once we have a React hook in place.

Otherwise, there's just some shuffling around to keep things organized! Here's a look at color interpolation in action:

191212_color_interpolation


__tests__/core/Vector_test.re
```diff
@@ -189,7 +189,7 @@ describe("Vector", () => {
       "should expose a function for linearly interpolating between two integer vectors",
       () =>
       Expect.(
-        expect(Vector.lerp(~acc=v1Int, ~target=v2Int, ~roundness))
+        expect(Vector.lerpV(~acc=v1Int, ~target=v2Int, ~roundness))
```
parkerziegler: Now that we have `lerp` being used more consistently as a pure mathematical operation on scalars, I figured we should isolate this vector version by appending `V`. I may do this to other vector functions to distinguish them down the road, but not a big deal since we likely won't even ship most of these as part of the public API 👍

__tests__/interpolate/Remap_number_test.re
```diff
@@ -0,0 +1,61 @@
+open Jest;
+
+let it = test;
+
+describe("Remap_number", () =>
```
parkerziegler: `remapf` is kind of like `d3-scale` for folks who are familiar. It answers the basic question:

> Given an input range and an output domain, where would a value in the input range fall in the output domain?

`normalizef` just does this with [0, 1] as the pre-configured output range.

src/core/Utils.re
```diff
@@ -1,28 +1,2 @@
 let foi = float_of_int;
-let iof = int_of_float;
-
-type scanfState = {
```
parkerziegler: I typically don't like to delete code until later, but I actually think this version of `scan` / `reduce` is pretty meh. It's also not currently in use, so I decided to scrap it for now with the idea that we may be able to improve it later if we need it.

src/interpolate/Lerp_color.rei
```diff
@@ -0,0 +1,11 @@
+type rgba('a) = {
+  r: 'a,
+  g: 'a,
+  b: 'a,
+  a: 'a,
+};
```
parkerziegler: This structure in Reason is called a **record**. The easiest mental model is probably JS objects, but they actually behave a bit differently. Records are immutable by default and _require_ a predefined type. They're quite memory efficient and safe, and a good choice for something like colors. BuckleScript now supports compiling them to pure JS objects but for awhile they were actually stored as arrays.

PR #8, opened 2019-12-15
Initial useGravity API.

parkerziegler:
This PR seems large, but again, lots of autogenerated code here, so the files I'd love to have you focus in on are:

- `src/hooks/useGravity.ts`
- `src/helpers/pairs.ts`

This PR represents the v1 of our `useGravity` hook 🎉 The API for the hook looks like this:

```typescript
const [props] = useGravity<HTMLDivElement>({
  from: { opacity: 0 },
  to: { opacity: 1 },
  config: {
    moverMass: 100000,
    attractorMass: 1000000000,
    r: 150
  }
});

return <div {...props} />
```

The core idea is that `props` right now is really just a `ref` we pass back to the consumer of a hook (thanks to @bmathews for proposing this). Once it gets spread on we're able to mutate that element's `style` properly directly, thus circumventing React state / re-rendering nightmares when trying to keep that 60fps target in sight.

The other big idea in this PR is the introduction of interpolators. The current flow I have in mind is this:

- Parse the `from` / `to` pair provided by the user to a normalized object structure of:

```typescript
{
  from: {
    property: 'opacity',
    value: 0,
  },
  to: {
    property: 'opacity',
    value: 1
  },
}
```

- Perform some light error checking to ensure they're animating the same property and that the `from` / `to` values are of the same type.
- If the `typeof value` is a number, we can do a normal linear interpolation between number values, using the `mover`'s distance from the `attractor` as the input range and the `from` / `to` values as the output range.
- If the `typeof value` is a string, we have a bit more work to do. We need to figure out if they handed us:
   - A color – this is handled by `normalizeColor` and is included in this PR. If so, use `remapColor` as the interpolator.
  - A string value with a unit – i.e. `"50%"`, `"10rem"`, etc.
  - A `transform` operation – i.e. `"scale(1.5)"`.
  - A `box-shadow` – i.e. `"box-shadow: 60px -16px teal;"`

These last three cases are a bit tricky, so I'm going to handle them in a follow up PR.

Overall I think this API feels pretty good, but I'd love to here thoughts from others on how this feels. There's considerably less magic to it than in `react-spring`, but that feels right to me at least for the time being. We can always scale up the complexity and sophistication as we need ✅


__tests__/helpers/pairs.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { parsePairs } from "../../src/helpers/pairs";
```
parkerziegler: We have support for tests in TS as well now 🎉 I'm thrilled with how nice the interop b/t Reason and TS has been 😄

__tests__/interpolate/Interpolate_color_test.re
```diff
@@ -0,0 +1,39 @@
...
+          Interpolate_color.lerpColorRGBA(~acc=a, ~target=b, ~roundness=0.5),
+        )
+        |> toEqual(Interpolate_color.{r: 50., g: 100., b: 127.5, a: 0.3})
+      );
+    })
+  );
+
+  describe("remapColor", () =>
+    it("should map an input numeric range to an output color domain", () => {
+      let a = Interpolate_color.{r: 0., g: 0., b: 0., a: 0.};
```
parkerziegler: To create a record in Reason its type definition needs to be brought into scope. That's why we locally open the `Interpolate_color` module here – it brings the `rgba` record type into scope so we can create an `rgba` record.

__tests__/interpolate/Interpolate_number_test.re
```diff
@@ -21,7 +32,7 @@ describe("Remap_number", () => {
     it("should handle ranges and domains with non-overlapping inputs", () =>
       Expect.(
         expect(
-          Remap_number.remapf(
+          Interpolate_number.remapf(
```
parkerziegler: I decided to keep all interpolation related functions in a single module `Interpolate_number`.

babel.config.js
```diff
@@ -7,6 +7,7 @@ module.exports = {
           node: "current"
         }
       }
-    ]
+    ],
+    "@babel/preset-typescript"
```
parkerziegler: This will transpile our TS! We didn't need this with Parcel, but we need it now to support Jest.

example/index.tsx
```diff
@@ -37,47 +29,23 @@ const App: React.FC = () => {
     }
   });
 
-  // Gravity force for the opacity demo.
-  gravity1D({
+  const [propsBackground] = useGravity<HTMLDivElement>({
+    from: { background: "rgba(100, 0, 100, 0.5)" },
+    to: { background: "rgba(100, 200, 200, 1)" },
```
parkerziegler: 🎉 We can interpolate colors! This _should_ also handle `rgb`, `hsl`, and HEX as well, but I'm going to write unit tests to make sure.

src/helpers/LICENSE
```diff
@@ -0,0 +1,29 @@
+BSD 3-Clause License
```
parkerziegler: We're using `normalize-css-color` but, in order to get proper type support, I copied the repo source over here and added type definitions / some ES6 sugar it was missing. It seemed from their license that we needed to include the LICENSE file in source. cc/ @ryan-roemer any thoughts here on what other steps we should take? Here's the original repo: https://github.com/react-native-community/normalize-css-color `react-spring` seems to do this as well although the license text is in file here: https://github.com/react-spring/react-spring/blob/master/src/shared/normalizeColors.ts

src/helpers/normalizeColor.ts
```diff
@@ -0,0 +1,367 @@
+/*
```
parkerziegler: No need to review here, but worth taking a look at all the cool strategies for normalizing colors that `normalize-css-color` uses. Lots of bitwise operators!

src/helpers/pairs.ts
```diff
@@ -0,0 +1,104 @@
...
+export const parsePairs = ({ from, to }: CSSPairs) => ({
+  from: parsePair(from),
+  to: parsePair(to)
+});
+
+type Interpolator<T> = (params: {
+  readonly range: [number, number];
+  readonly domain: [T, T];
+  readonly value: number;
+}) => T;
```
parkerziegler: All interpolator functions _must_ match this shape. Having a normalized shape for interpolators like this helps to keep our functions much cleaner.

src/helpers/pairs.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    return {
+      interpolator: remapColor,
+      property: fromProperty,
+      values: {
+        from: rgba(normalizedFrom !== null ? normalizedFrom : fromValue),
+        to: rgba(normalizedTo !== null ? normalizedTo : fromValue)
+      }
+    };
+  }
```
parkerziegler: Currently the only string type we support is colors. I'll tackle string values like `"10px"`, `"40rem"`, etc. in a future PR.

src/hooks/useGravity.ts
```diff
@@ -0,0 +1,44 @@
...
+}: UseGravityArgs) => {
+  /**
+   * Store a ref to the DOM element we'll be attaching to.
+   * A user will spread this ref onto their own element, which
+   * is what allows us to directly update the style property
+   * without triggering rerenders.
+   */
+  const ref = React.useRef<T>(null);
+
+  const [stop] = React.useMemo(
```
parkerziegler: So @bmathews made a great point. Our components might be re-rendering a bunch, but we don't want that behavior to restart our animation. This is my first naive attempt at controlling that behavior, but I'm not convinced it'll work. If anyone has thoughts on how to preserve animation state across re-renders I'd be very interested!

kitten:
This is looking really great! Got to dig in some more but before I do I’ll just drop some questions here 🙌

- Are you supporting time based continuation where the animation state could be preserved in React and passed back in, in case the memo disappears?
- Are we handling “exotic” animation properties like stylefire does, where the properties that are animated are automatically interpolated? (Presumably for colors, yes, but what about transforms?)

parkerziegler:
> Are you supporting time based continuation where the animation state could be preserved in React and passed back in, in case the memo disappears?

I think I'd like to, but I'm not totally sure of how to implement this 😅 If you have any good reading or resources on this, I'd love to take a look. This definitely seems like a stretch goal for the six weeks for me, but it is something that I think would make the lib awesome.

> Are we handling “exotic” animation properties like stylefire does, where the properties that are animated are automatically interpolated? (Presumably for colors, yes, but what about transforms?)

Ideally, yes. Part of the goal of this PR was getting a normalized shape for those interpolators in place (i.e. accept an input `range`, output `domain`, and the `value` we want to interpolate) and then leaving the implementation details up to the specific interpolator. A `transform` interpolator would be absolutely ideal, supporting things like `translate`, `rotate`, `scale`. My goal is to get these working incrementally and continue to fill them in as we can. The next PR should show a bit more my thoughts on architecting this. And who knows, I could even see adopting something like `stylefire` under the hood to help do this for us.

PR #9, opened 2019-12-16
Add support for interpolation of CSS units.

parkerziegler:
This PR adds support for interpolation of values like "45px", "10rem", "50%", etc. which is useful for animating positioning properties like `top`, `left`, `bottom`, `right`, `height`, `width`, `padding`, `margin`, `maxHeight`, `maxWidth`, `borderWidth`, etc. Basically, a lot of things!

The basics of how we do this are:

1. Split a string into its numeric and unit parts. We do this by calling `parseFloat` first to obtain the numeric portion, and then treating the rest of the string as the unit (a clever trick from Popmotion :chefs-kiss:).
2. We then validate the unit matches one of `px`, `em`, `rem`, `vh`, `vw`, and `%`. If I'm missing any here please let me know (maybe `pt`?)
3. We then interpolate the numeric portion using regular linear interpolation.
4. We then reconstitute the string and hand it back to the calling function.

This is pretty sweet! There's one drawback. We can't interpolate if they hand us strings that are of different unit types, i.e. `from: 10px` and `to: 5rem`. If a user did that now, it would use the unit of the `from` property, so in the above it'd interpolate from `10px` to `5px`. Feels like an optimization we can tackle more downstream.

In general, I took a lot of inspiration from https://github.com/Popmotion/popmotion/tree/master/packages/style-value-types, which is just absolutely :dope:. However, I reimplemented things in Reason b/c our other interpolators are all written in Reason and I want to keep that experience consistent for the time being.

Here's the `left` property being animated with this approach now:

**Code**
```typescript
const [props] = useGravity<HTMLDivElement>({
  from: { left: '10px' },
  to: { left: '160px' },
  config: {
    moverMass: 100000,
    attractorMass: 1000000000,
    r: 150,
  },
});

return <div {...props} />;
```

**Output**

191216_use_gravity_left


__tests__/helpers/pairs.spec.ts
```diff
@@ -1,13 +1,99 @@
...
+    });
+
+    it('should infer a from / to pair of type string that matches a color and return a color interpolator', () => {
+      const pair = {
+        from: { background: '#c86432bf' }, // rgba(200, 100, 50, 0.75)
+        to: { background: '#64c84b' }, // rgba(100, 200, 75, 1)
+      };
+      const { interpolator } = (getInterpolatorForPair(pair) as unknown) as {
+        interpolator: Interpolator<RGBA<number>, string>;
+      };
```
parkerziegler: In truly lame fashion, TS doesn't seem to be properly inferring the overloads on this function based on how it's used – good ol' ISTS. If anyone is an expert in getting TS overloads to play nicely, definitely hit me up!

src/core/Utils.re
```diff
@@ -1,2 +1,3 @@
 let foi = float_of_int;
-let iof = int_of_float;
\ No newline at end of file
+let iof = int_of_float;
+let sof = Js.Float.toString;
```
parkerziegler: Reason does have `string_of_float` but it apparently is quite poor at producing accurate strings, so the compiler recommends `Js.Float.toString`.

src/helpers/normalizeColor.ts
```diff
@@ -163,6 +163,95 @@ const colorNames = {
 
 type ColorName = keyof typeof colorNames;
 
+const NUMBER = '[-+]?\\d*\\.?\\d+';
```
parkerziegler: Just reordering things in this file such that they're defined before being used, no need to review.

src/helpers/pairs.ts
```diff
@@ -40,6 +46,9 @@ export function getInterpolatorForPair(
 export function getInterpolatorForPair(
   pairs: CSSPairs
 ): InterpolatedResult<RGBA<number>, string>;
+export function getInterpolatorForPair(
+  pairs: CSSPairs
+): InterpolatedResult<string, string>;
```
parkerziegler: A new overload. In the case of values like "10px", the interpolator should return a string value as well with the same unit.

src/interpolate/Interpolate_unit.re
```diff
@@ -0,0 +1,33 @@
+type cssUnit = {
+  num: float,
+  unit: Js.Nullable.t(string),
+};
+
+[@bs.deriving jsConverter]
+type measurement = [ | `px | `em | `rem | `vw | `vh | [@bs.as "%"] `pct];
```
parkerziegler: I talked about this a bit here: https://formidablelabs.slack.com/archives/CR02GE56Y/p1576533160023200. Basically, when we get a string from JS-land we call `measurementFromJs`, which will attempt to convert it to one of these polymorphic variants. If it finds a match, we'll get `Some`, otherwise `None`. Then on L16-17 below we convert it back. This is what allows us to return `null` for `unit` if the passed `unit` doesn't match one of our polymorphic variants. Happy to go more in depth on this if folks want. Here's the associated documentation: https://bucklescript.github.io/docs/en/generate-converters-accessors#convert-between-js-string-enum-and-bs-polymorphic-variant

PR #11, opened 2019-12-18
Add support for interpolating transforms.

parkerziegler:
This PR adds support for interpolating `transform`s 🔥 After digging in a little bit, I soon realized this is a pretty sizeable task. `transform` is a pretty extensive CSS property, supporting `translate`, `rotate`, `skew`, `scale`, and `matrix` along with their `-X`, `-Y` and `-Z` counterparts. Oh, and something called `perspective` I'd never used before 😂  There's also the trickiness that some of these accept multiple arguments, i.e. `translate(x, y)` or `scale(x, y)`. Add to that the fact that `transform` can accept _multiple_ of these add a time (i.e. `transform: translate(x, y) scale(z)`) and we get a pretty challenging string to parse apart and reconstitute.

For this early work, I decided to simplify things to work just with a single `transform` with a single property. For example, the code currently would work for `scale(1.2)`, `translateX(5px)`, `rotateZ(10rad)`, but **wouldn't** support `translate(10px, 60px)`, `scale(1.2, 5.6)` or `scaleX(1.2) rotateY(0.5)`. I figure since we have the core in place, adding those on later will be a matter of improving the algorithms and better abstracting the parsing behavior. It could also be great work for newcomers to Reason to get involved with the project, since it's a very algorithmic problem.

I'm going to add examples showing these, but I want to get Storybook setup first!


__tests__/interpolate/Interpolate_transform_test.re
```diff
@@ -0,0 +1,153 @@
...
+               Interpolate_transform.{
+                 transformProperty: Js.Nullable.return("skewZ"),
+                 transform: "1.5rad",
+               },
+               Interpolate_transform.{
+                 transformProperty: Js.Nullable.return("perspective"),
+                 transform: "10px",
+               },
+             ))
+        );
```
parkerziegler: Since `bs-jest` allows at max one assertion per test, the canonical way to test a lot of things at once is to place them all in a tuple. In this case, a five-dimensional tuple (it has five elements)!

src/interpolate/Interpolate_transform.re
```diff
@@ -0,0 +1,76 @@
...
+  | `rotate
+  | `rotateX
+  | `rotateY
+  | `rotateZ
+  | `scale
+  | `scaleX
+  | `scaleY
+  | `scaleZ
+  | `perspective
+];
```
parkerziegler: Using our polymorphic variants once again! This helps us to ensure that the user has actually passed us a valid property to use with `transform`. I'm leaving `matrix` out for now – too complicated for a Wednesday 😂

src/interpolate/Interpolate_transform.re
```diff
@@ -0,0 +1,76 @@
...
+  | idx =>
+    /* Decompose the translate string into the translate property, i.e. translateX, skew, etc.
+       Then, extract the value, i.e. 20px, 50deg, so we can begin to interpolate it. */
+    let property =
+      Js.String.substring(~from=0, ~to_=idx, transform)
+      ->transformPropertiesFromJs;
+    let value =
+      Js.String.substring(
+        ~from=idx + 1,
+        ~to_=String.index_from(transform, idx + 1, ')'),
```
parkerziegler: So here we're basically hedging a bet that a user is giving us a string that has both a closing and opening parenthesis. We _could_ be even more type safe here, but it feels a bit overkill right now. Hopefully people aren't thinking that this lib should handle `transform: translateX(4px` (no closing parenthesis).

src/interpolate/Interpolate_transform.re
```diff
@@ -0,0 +1,76 @@
...
+      );
+
+    switch (property) {
+    | Some(p) => {
+        transform: value,
+        transformProperty: Js.Nullable.return(p->transformPropertiesToJs),
+      }
+    | None => {transform: value, transformProperty: Js.Nullable.null}
+    };
+  | exception Not_found => {transform, transformProperty: Js.Nullable.null}
```
parkerziegler: This is how you handle `exception`s in Reason. If we don't find an opening parenthesis in the string handed us by the user, we'll return `null` for the `transformProperty`. This will cause our `getInterpolatorForPair` function to say, "Alright, I bet this isn't a valid CSS transform."

src/interpolate/Interpolate_transform.re
```diff
@@ -0,0 +1,76 @@
...
+    Interpolate_number.lerpf(~acc=dlNum, ~target=dhNum, ~roundness=progress);
+
+  switch (
+    dlTransformProperty->Js.Nullable.toOption,
+    dlUnit->Js.Nullable.toOption,
+  ) {
+  | (Some(tp), Some(u)) => tp ++ "(" ++ Utils.sof(output) ++ u ++ ")"
+  | (Some(tp), None) => tp ++ "(" ++ Utils.sof(output) ++ ")"
+  | (None, Some(u)) => Utils.sof(output) ++ u
+  | (None, None) => ""
```
parkerziegler: Pattern matching ftw! If we detected both a valid `transform` property and a valid unit property, we'll rebuild the string (an example being `translateX(20px)`). If we detected a valid `transform` property and a unitless number, we'll still rebuild the string (an example being `scale(1.5)`. If we didn't detect a `transform` property, we'll still do out best to return a united output (i.e. `20px`). Otherwise, we'll just return an empty string. This last condition really shouldn't get hit.

PR #12, opened 2019-12-19
Migrate from Parcel to Storybook for prototyping. Add onComplete callback.

parkerziegler:
This PR migrates our example environment away from Parcel to [Storybook](https://storybook.js.org/). The benefit of this migration is two-fold. First, our examples can be better isolated, which I think will both make them more easily discoverable and debug-able. Second, with the addition of Storybook's knobs [addons](https://github.com/storybookjs/storybook/tree/next/addons/knobs), we can tweak parameters of the animation directly in Storybook and see changes take effect. This should give contributors to the library a great place to test their changes.

191219_storybook

The other core addition to this PR is the `onComplete` callback that gets passed to `gravity1D`. I was noticing that our `from` / `to` animations weren't actually reaching `to`. You'd have to get exceedingly lucky, where the force applied in the last frame before overshooting placed the `mover` at _exactly_ the same location as the `attractor`. To get around this, we pass an `onComplete` callback to `gravity1D`. Once the `isOvershooting` condition is reached, we run `onComplete` and `stop` to halt the frame loop. Currently, `onComplete` just mutates the `style` of our animating element to match `config.to` exactly. If other folks have ideas here I'd love to hear them! This is a very Rx'y pattern but I think works out pretty well!

One thing that's a bit of a bummer – Storybook appears to mount all components in all tabs on load, so the animations run immediately 😬 This means if you switch tabs, all animations will have already run and you'll need to reload to see them take effect. It may be nice down the road to provide explicit `start` and `stop` buttons for the user, but not a huge deal now while I'm prototyping.


.storybook/webpack.config.js
```diff
@@ -0,0 +1,11 @@
...
+  config.module.rules.push({
+    test: /\.(ts|tsx)$/,
+    loader: require.resolve('babel-loader'),
+    options: {
+      presets: [['react-app', { flow: false, typescript: true }]],
+    },
+  });
+  config.resolve.extensions.push('.ts', '.tsx');
+  return config;
+};
```
parkerziegler: Custom `webpack` configuration for Storybook. This just runs `babel-loader` such that we strip off all the type information in our `.stories.tsx` files.

stories/useGravity.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+  title: 'Gravity',
+  decorators: [withKnobs],
+};
+
+export const GravityOpacity: React.FC = () => {
+  const [props] = useGravity<HTMLDivElement>({
+    from: { opacity: 0 },
+    to: { opacity: 1 },
+    config: {
+      moverMass: number('moverMass', 100000),
```
parkerziegler: The `number` piece is what configures this parameter as a knob that can get tweaked by a user running Storybook.

stories/useGravity2D.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export const Gravity2D: React.FC = () => {
+  const [center, setCenter] = React.useState<Vector<number>>([0, 0]);
+
+  React.useLayoutEffect(() => {
+    const root = document.getElementById('root');
+    if (root) {
+      setCenter([root.clientWidth / 2, root.clientHeight / 2]);
+    }
+  }, []);
```
parkerziegler: This was an interesting little problem that led me to read a bit more about `useLayoutEffect`. Basically, I want to derive the center point of the `#root` node to center the `attractor`. We can't use `document.body` b/c Storybook purposefully gives it no height 🤔

image

I haven't yet encountered a good case for `useLayoutEffect` in the wild, but this seemed like one – we basically want a sync render so the user doesn't get a flicker of the `attractor` being in the upper left corner ([0, 0]) and then suddenly changing location to the center.

PR #16, opened 2019-12-30
Add initial useFriction prototype.

parkerziegler:
This PR adds support for our second force - friction! We're starting with the classic `from` / `to` style animations in a single dimension before getting into more complex animations like moving down surfaces or even transitioning from one surface to another.

The most important piece in this code is the stopping condition I chose to determine when the animation has finished. You may remember from `useGravity` that the stopping condition is hit when the `mover` has reached the `attractor`. This felt like a fairly straightforward choice that matched my expectations of movement. For friction, there isn't as clear a stopping condition.

The one I went with here is that the moving body has reached rest, i.e. hit a velocity of 0. This means that the length of the animation is directly proportional to the coefficient of friction (`mu`) and the `normal` force, and inversely proportional to `velocity`. If a body is moving really fast, it will take a long time to reach a velocity of 0. If a body is moving on a rough surface (high `mu`) or is heavier (high `normal`), it will slow down more rapidly than if its on a smooth surface (low `mu`) or is lighter (low `normal`).

To be frank, I'm not sure this is the right decision. As you increase velocity, animations take _longer_, which is a bit confusing. I'm thinking a better solution would be to provide a `distance` parameter that a user can adjust. The length of the animation would then be related to the amount of time it takes for a moving object to traverse that distance across the frictive surface, or to come to rest if it can't cross the distance. This would make the animation behave a bit more like users might expect (high initial velocity results in shorter animation). I'm going to keep playing with this, but wanted to get first thoughts now!

**Edit**

After some thinking and planning, I think I arrived at a nice solution for the problem of a stopping condition! The issue in the first iteration was that we were using our velocity value as our animation value, i.e. we'd interpolate changes to our CSS property according to changes in velocity. However, acceleration is constant in a pure friction scenario, meaning velocity decreases linearly over time. To achieve a nice, smooth animation effect, we actually need to use position.

The solution I arrived at involves using [kinematic equations](https://www.khanacademy.org/science/physics/one-dimensional-motion/kinematic-formulas/a/what-are-the-kinematic-formulas). Basically, the idea is we derive *ahead of time* how far an object can travel over a rough surface using a derivation of the fourth kinematic equation:

∆x = v_initial^2 / -2 * a_friction

This then sets up the `range` of of our interpolation, and we can just watch the `position` value over time to interpolate our CSS properties!

191241_use_friction


src/animation/friction.ts
```diff
@@ -0,0 +1,79 @@
...
+  normal?: number
+) => {
+  const force = frictionForceV({ mu, normal, velocity: entity.velocity });
+
+  return applyForce({ force, entity });
+};
+
+export const friction1D = (params: Friction1DParams) => {
+  let state: FrictionState = {
+    mass: 1,
```
parkerziegler: Since we're making the normal force a scalar constant, `mass` can just be 1 here. On a 2d surface we'd derive the normal force using the angle between the direction of the normal force and the gravitational force. But for the purpose of these 1d animations, we're assuming a flat surface where the normal force and gravitational force cancel one another out.

src/animation/friction.ts
```diff
@@ -0,0 +1,79 @@
...
+    for (let i = 0; i < steps; i++) {
+      state = applyForceForStep(state, params.config.mu, params.config.normal);
+    }
+
+    /**
+     * Conditions for stopping the physics animation. For single animations with
+     * a discrete from / to pair, we want to stop the animation once the moving
+     * object has a velocity of 0 (has come to rest).
+     */
+    if (state.velocity[0] <= 0) {
```
parkerziegler: Here's the stopping condition where we ensure the velocity has reached 0. I'm thinking a distance based calculation may give a better feel for the animations and allow the API to operate a bit more like users might expect.

src/hooks/useFriction.ts
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+
+import { CSSPairs, getInterpolatorForPair } from '../helpers/pairs';
+import { Friction1DParams, friction1D } from '../animation/friction';
+
+type UseFrictionArgs = CSSPairs &
+  Omit<Friction1DParams, 'onUpdate' | 'onComplete'>;
+
+export const useFriction = <M extends HTMLElement>({
```
parkerziegler: In truly great fashion, our `useFriction` hook is nearly _identical_ to our `useGravity` hook, with just a few different config parameters and calling of a different physics calculation (`friction1D` vs. `gravity1D`). 🎉 Love when it's this easy!

PR #17, opened 2020-01-02
Add support for immediate and delay properties. Move start and stop into the controller.

parkerziegler:
This PR adds support for what I'm calling the "controller" API. Basically, a `controller` allows a user to start and stop their animations arbitrarily. It works this:

```typescript
const [props, controller] = useFriction<HTMLDivElement>({
  from: { transform: 'translateX(0px) translate(-50%, -50%)' },
  to: { transform: 'translateX(300px) translate(-50%, -50%)' },
  config: {
    mu: 0.5,
    mass: 300,
    initialVelocity: 10,
  },
  // By default, all animations will run immediately on component mount.
  // This instructs the hook to wait until controller.start has been called.
  immediate: false,
});

return (
  <>
    // In onClick we call `controller.start` to begin the animation. You also have access to controller.stop.
    <button className="button" onClick={controller.start}>
      Run Animation
    </button>
    <div className="mover mover--translate" {...props} />
  </>
);
```

This is great, because we can now arbitrarily start and stop animations in response to events! I did decide to default the `immediate` option to `true`, because I imagine most users will expect the animation to run immediately with no additional configuration.

This PR also adds a `delay` option to the configuration. `delay` queues an animation to run after a specified timeout. Currently, there's only support for delaying an animation that doesn't wait for a user event to actually initiate it. I plan to fix this in an upcoming PR, but I want to get some feedback on this approach first.

Finally, I added support for interpolating multiple CSS properties simultaneously 🎉 This means users can now do things like:

```typescript
from: {
   opacity: 0,
   transform: 'scale(0) rotate(0deg)'
},
to: {
  opacity: 1,
  transform: 'scale(1) rotate(180deg)'
}
```

We're now able to do some pretty nifty stuff!

200102_friction_toggle

### DIFF
--- a/src/animation/friction.ts
+++ b/src/animation/friction.ts
@@ -1,15 +1,13 @@
 import { rAF } from '../rAF';
 import { Entity, applyForce, frictionForceV } from '../forces';
-import { VectorSetter, Controller } from './types';
+import { Listener, AnimationParams, AnimationInitializer } from './types';
 
-export interface Friction1DParams {
+export interface Friction1DParams extends AnimationParams {
   config: {
     mu: number;
     mass: number;
     initialVelocity: number;
   };
-  onUpdate: VectorSetter;
-  onComplete: () => void;
 }
 
 interface FrictionState {
@@ -41,18 +39,17 @@ const applyFrictionForceForStep = (
  */
 export const friction1D = (
   params: Friction1DParams
-): { controller: Controller } => {
+): { controller: AnimationInitializer } => {
   const state: FrictionState = {
     mover: {
       mass: params.config.mass,
       acceleration: [0, 0],
-      velocity: [params.config.initialVelocity, 0],
+      velocity: [params.config.initialVelocity * 1000, 0],
       position: [0, 0],
     },
   };
 
-  const { start } = rAF();
-  const { stop } = start((timestamp, lastFrame, stop) => {
+  const listener: Listener = (timestamp, lastFrame, stop) => {
     /**
      * Determine the number of milliseconds elapsed between the current frame
      * and the last frame. If more than four frames have been dropped, assuming
@@ -92,7 +89,10 @@ export const friction1D = (
         position: state.mover.position,
       });
     }
-  });
+  };
 
-  return { controller: { start, stop } };
+  const { start } = rAF();
+  const runAnimation = () => start(listener);
+
+  return { controller: { start: runAnimation } };
 };
--- a/src/animation/friction.ts
+++ b/src/animation/friction.ts
@@ -44,6 +44,7 @@ export const friction1D = (
     mover: {
       mass: params.config.mass,
       acceleration: [0, 0],
+      // Initial velocity is provided in m / ms. Multiply by 1000 to derive m / s.
       velocity: [params.config.initialVelocity * 1000, 0],
       position: [0, 0],
     },
--- a/src/helpers/pairs.ts
+++ b/src/helpers/pairs.ts
@@ -149,3 +149,12 @@ export function getInterpolatorForPair({ from, to }: CSSPairs) {
     },
   };
 }
+
+export function getInterpolatorsForPairs({ from, to }: CSSPairs) {
+  return Object.keys(from).map(key => {
+    const f = { [key]: (from as { [cssProperty: string]: any })[key] };
+    const t = { [key]: (to as { [cssProperty: string]: any })[key] };
+
+    return getInterpolatorForPair({ from: f, to: t });
+  });
+}
--- a/src/hooks/useFriction.ts
+++ b/src/hooks/useFriction.ts
@@ -1,16 +1,18 @@
 import React from 'react';
 
-import { CSSPairs, getInterpolatorForPair } from '../helpers/pairs';
+import { CSSPairs, getInterpolatorsForPairs } from '../helpers/pairs';
 import { Friction1DParams, friction1D, Controller } from '../animation';
 import { getMaxDistanceFriction } from '../forces';
 
 type UseFrictionArgs = CSSPairs &
-  Omit<Friction1DParams, 'onUpdate' | 'onComplete'>;
+  Omit<Friction1DParams, 'onComplete' | 'onUpdate'>;
 
 export const useFriction = <M extends HTMLElement>({
   from,
   to,
   config,
+  immediate = true,
+  delay,
 }: UseFrictionArgs): [
   { ref: React.MutableRefObject<M | null> },
   Controller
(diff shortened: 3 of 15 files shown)


src/animation/friction.ts
```diff
@@ -41,18 +39,17 @@ const applyFrictionForceForStep = (
...
 export const friction1D = (
   params: Friction1DParams
-): { controller: Controller } => {
+): { controller: AnimationInitializer } => {
   const state: FrictionState = {
     mover: {
       mass: params.config.mass,
       acceleration: [0, 0],
-      velocity: [params.config.initialVelocity, 0],
+      velocity: [params.config.initialVelocity * 1000, 0],
```
parkerziegler: In order to get realistic motion previously, our velocity values had to be _super high_. This is because we're calculating the force every millisecond, but our equation assumes seconds as the time unit. Multiplying by 1000 here helps to convert everything to the proper reference units for deriving force.

src/animation/friction.ts
```diff
@@ -92,7 +89,10 @@ export const friction1D = (
...
       });
     }
-  });
+  };
+
+  const { start } = rAF();
+  const runAnimation = () => start(listener);
 
-  return { controller: { start, stop } };
+  return { controller: { start: runAnimation } };
```
parkerziegler: Rather than calling `start` inside of this function, we now return a function that calls `start` with the `listener` already bound. This is what allows us to attach `start` to the `controller` for our users to use.

src/helpers/pairs.ts
```diff
@@ -149,3 +149,12 @@ export function getInterpolatorForPair({ from, to }: CSSPairs) {
...
 }
+
+export function getInterpolatorsForPairs({ from, to }: CSSPairs) {
+  return Object.keys(from).map(key => {
+    const f = { [key]: (from as { [cssProperty: string]: any })[key] };
+    const t = { [key]: (to as { [cssProperty: string]: any })[key] };
+
+    return getInterpolatorForPair({ from: f, to: t });
+  });
+}
```
parkerziegler: This allows for simultaneous interpolation of multiple CSS properties! For example:

```typescript
from: {
   opacity: 0,
   transform: 'scale(0)'
},
to: {
  opacity: 1,
  transform: 'scale(1)'
}
```

src/hooks/useFriction.ts
```diff
@@ -1,16 +1,18 @@
...
 
 type UseFrictionArgs = CSSPairs &
-  Omit<Friction1DParams, 'onUpdate' | 'onComplete'>;
+  Omit<Friction1DParams, 'onComplete' | 'onUpdate'>;
 
 export const useFriction = <M extends HTMLElement>({
   from,
   to,
   config,
+  immediate = true,
```
parkerziegler: I just added these to `useFriction` thus far. I wanted to test things out here before going whole hog and applying them to `useGravity`.

src/hooks/useFriction.ts
```diff
@@ -23,42 +25,72 @@ export const useFriction = <M extends HTMLElement>({
...
+      timerId = setTimeout(() => {
+        const { stop } = controller.start();
+        controllerRef.current.stop = stop;
+      }, delay);
+    }
+
+    return () => {
+      timerId && clearTimeout(timerId);
+    };
+  }, [immediate, delay, controller]);
```
parkerziegler: `useLayoutEffect` is important here. Once we call `controller.start` we're going to want to synchronously mutate the DOM via the `ref` here. We want this mutation to take effect before the browser actually paints, hence why we use `useLayoutEffect` here.

stories/useFriction.stories.tsx
```diff
@@ -3,49 +3,94 @@ import { withKnobs, number } from '@storybook/addon-knobs';
...
+      opacity: toggle ? 0 : 1,
+      transform: toggle
+        ? 'scale(0) rotate(0deg) translate(-50%, -50%)'
+        : 'scale(1) rotate(180deg) translate(-50%, -50%)',
+    },
+    to: {
+      opacity: toggle ? 1 : 0,
+      transform: toggle
+        ? 'scale(1) rotate(180deg) translate(-50%, -50%)'
+        : 'scale(0) rotate(0deg) translate(-50%, -50%)',
```
parkerziegler: Changing `from` / `to` values will cause the animation to re-run, which is exactly what we want. This allows users to hook these values up to, say, a toggle, or an effect that runs the animations on an interval by changing the `from` / `to` values.

stories/useFriction.stories.tsx
```diff
@@ -3,49 +3,94 @@ import { withKnobs, number } from '@storybook/addon-knobs';
...
-      initialVelocity: number('velocity', 1000),
+      initialVelocity: number('velocity', 10),
     },
+    immediate: false,
   });
 
-  return <div className="mover mover--translate" {...props} />;
+  return (
+    <>
+      <button className="button" onClick={controller.start}>
```
parkerziegler: The first use of `controller` – wiring it to a button click. The animation won't run until `controller.start` is called. Notice the `immediate: false` above. I'd love any thoughts on this API and whether or not it feels right // intuitive.

stories/useFriction.stories.tsx
```diff
@@ -3,49 +3,94 @@ import { withKnobs, number } from '@storybook/addon-knobs';
...
+      transform: 'scale(1.5) rotate(720deg) translate(-50%, -50%)',
+    },
     config: {
-      mu: number('mu', 0.05),
+      mu: number('mu', 0.5),
       mass: number('mass', 300),
-      initialVelocity: number('velocity', 1000),
+      initialVelocity: number('velocity', 10),
     },
+    delay: 2000,
```
parkerziegler: The `delay` option! This animation will wait 2s before starting.

src/animation/friction.ts
kitten: I can’t quite see where the other part of the velocity code is but if you’re using milliseconds it’s probably worth converting to seconds there in the formula? Or are milliseconds already established as the base time frame there? Kind of unexpected imho judging from a metric standpoint 😅

parkerziegler: Here's the other place where it's used: https://github.com/FormidableLabs/renature/pull/17/files#diff-3dcc94f37ffa7fc7c99d5f790af6589fR45 Basically, we use the third kinematic equation to answer the question:

> If an object is moving at v_0 m/s on a rough surface, how far will it travel by the time it has reached velocity of 0?

For that formula, the base time frame we're using is `ms`. But you're right, a better solution would be to do the conversion there and establish the base time frame as seconds. The challenge too is that we're applying the force once every `ms` (roughly, assuming no dropped frames). So just some work for me to do to reconcile the units a bit more cleanly. Good shout.

stories/useFriction.stories.tsx
littlemilkstudio: Yea, I think this totally works. 

However, I feel like a lot of it feeling a little funky is the fact that if you think about these transitions in terms of a finite automata, and you have `from` and `to`, It's almost as if your describing a the transition in terms of the current vertex as opposed to the possible state nodes it could be at and letting your physics handle the transition between the nodes.

I think another way of looking at this is you have two states `stateOne` & `stateTwo`. what you wanna do `onClick` is transition the current state _from_ `stateOne` _to_ `stateTwo` and let the `Friction` handle that interpolation. 

With the current API, you basically have to keep a mental model of the whole automata `stateOne <-> stateTwo`. you then describe a vertex with `from` & `to` (in this example you're describing the vertex: `stateOne -> stateTwo`). Then it becomes an implicit state transition by saying "execute the vertex" as opposed to an explicit state transition of "update to the new state".

I think if the api where altered the logical equivalent of what you have here wouldn't look as funky:
```
const [props, controller] = useFriction<HTMLDivElement>({
    states: {
      one: { transform: 'translateX(0px) translate(-50%, -50%)' },
      two: { transform: 'translateX(300px) translate(-50%, -50%)' },
    }
    initialState: 'one'
    config: {
      mu: number('mu', 0.05),
      mu: number('mu', 0.5),
      mass: number('mass', 300),
      initialVelocity: number('velocity', 1000),
      initialVelocity: number('velocity', 10),
    }
  });
...
   <button className="button" onClick={() => controller.setState('two')}>
```

I'm not saying you have to change your API by any means, but just saying that this is what's going on and how to reason about what's logically happening under the hood with the current API design. At least I think 🤷‍♂

Hopefully this helps. let me know if I'm completely off the mark and misunderstanding something too

parkerziegler: I love this description Max! It actually is really helpful to think about these animations in terms of states. In that sense, `initialState` of something like `init` or `standby` could be a really nice way to convey that the animation isn't yet running, but is _ready_ to run when executed. I don't think I'll necessarily be able to get to this by the time the Fellowship ends, but this could be a really excellent API, especially for things like sequencing animations. Thanks for the explanation, I'll make a longer term issue for this so we can track this alternative API design, because I think it's really cool and I would like to at least explore it.

littlemilkstudio: Dope. Yea, I don't think there's anything wrong with the current api describing the expected vertex to execute, but just another lens to view it through that could help with naming and other api decisions built on top of the current abstraction in the future.

Keep up the good work man. you're killing it

PR #18, opened 2020-01-06
Add support for a useFluidResistance hook.

parkerziegler:
This PR adds support for a `useFluidResistance` hook for modeling the drag force in a fluid. Drag is a dissipative force just like friction, acting in the opposite direction of velocity. Here's a look at the API:

```typescript
const [props] = useFluidResistance<HTMLDivElement>({
  from: {
    transform: 'translateY(-200%) translate(-50%, -50%)',
  },
  to: {
    transform: 'translateY(200%) translate(-50%, -50%)',
  },
  config: {
    mass: 250, // Mass of the object moving through the fluid.
    rho: 1, // Density of the fluid in kg/m^3.
    area: 20, // Frontal area of the moving object in m^2.
    cDrag: 0.1, // Coefficient of a drag, a unitless scalar.
    settle: true // Simulate a settling effect.
  },
});
```

and the result:

200106_useFluidResistance_seattle

Deciding on a stopping condition for this particular simulation was quite tricky. The simulation involves dropping an object from rest into a fluid and modeling its downward movement through the fluid as it's decelerated by the drag force. I decided to opt for [terminal velocity](https://en.wikipedia.org/wiki/Terminal_velocity) as the stopping condition. At terminal velocity, the drag force is equal to the gravitational force and an object achieves an acceleration of 0 and a constant velocity. You can read more about the math I'm using here: https://www.notion.so/formidable/Research-terminal-velocity-with-fluid-dynamics-98fd117953c740ebab7b430eead6816e

The idea behind the `settle` parameter is to give the effect of an element hitting a hard surface. If `settle` is set to `true`, we'll scale down the `velocity` and reverse its direction when the object achieves terminal velocity. We may want to consider other ways to do this, or write an exponential decay function to better simulate settling, but this is the naive way of doing it for the time being.

200106_useFluidResistance_settle_rotate


src/animation/fluidResistance.ts
```diff
@@ -0,0 +1,139 @@
...
+    rho: config.rho,
+    area: config.area,
+    velocity: mover.velocity,
+    cDrag: config.cDrag,
+  });
+
+  const gravitationalForce: Vector<number> = [0, mover.mass * gE];
+  const force = multf({
+    v: addf({ v1: dragForce, v2: gravitationalForce }),
+    s: 0.001,
```
parkerziegler: This is something I've been missing in other equations that I think we'll need to backport. Basically, our force vector is calculated using seconds, but is being _applied_ roughly once every ms. Therefore, we need to scale the vector down (multiply by 0.001) to achieve the force that would be applied every ms.

src/animation/fluidResistance.ts
```diff
@@ -0,0 +1,135 @@
...
+  const state: FluidResistanceState = {
+    mover: {
+      mass: config.mass,
+      acceleration: [0, 0],
+      velocity: [0, 0],
+      position: [0, 0],
+    },
+  };
+
+  const tvPosition = getFluidPositionAtTerminalVelocity(config);
```
parkerziegler: The position at which our moving object will achieve terminal velocity. We pre-calculate this and use it as the stopping condition for the animation.

src/animation/fluidResistance.ts
```diff
@@ -0,0 +1,135 @@
...
+    // Apply the drag force once for each step.
+    for (let i = 0; i < steps; i++) {
+      // If applying a settle effect, reverse and scale down the mover's velocity.
+      if (config.settle && state.mover.position[1] >= tvPosition) {
+        state.mover = {
+          ...state.mover,
+          velocity: multf({ v: state.mover.velocity, s: -0.75 }),
+          position: [0, tvPosition],
+        };
+      }
```
parkerziegler: Here's how we achieve the settle effect. Basically, we reverse the velocity vector of the mover and scale it down a little bit. We could consider exposing `s` here as a user parameter so they can set how much they want to scale the velocity down by. 0.75 is rather arbitrary here and just represents what "felt good".

src/animation/fluidResistance.ts
mariano-formidable: Good catch with this bug, I think you're correct, all the previous `applyForce` calls are being called every millisecond but applying 1 full second worth of acceleration & velocity :scream:

However, I don't think scaling down the force is applying the math correctly throughout the `applyForce` function.  I'll try and explain why I think so below:

```
Below are what I believe are the methods to calculate acceleration, velocity, and distance:

acc = F/m
vel = currVel + (acc * t)
pos = currPos + (vel * t)

In the current implementation of applyForce, you are _not_ using t.  Which implies that t = 1.  

So the formulas currently being used are:

acc = F/m
vel = currVel + acc ==> currVel + (acc * 1)
pos = currPos + vel ===> currPos + (vel * 1)

Your fix introduces a t = 0.001 multiplication at the F (force) level, but I don't think it propagates down correctly to the position.  

Here's my logic, by multiplying F by a t of 0.001, the applyForce formula updates as follows:

accWithT = (F * t)/m ==> (F/m) * t ==> acc * t (which is just velocity)
velWithT =  currVel + accWithT ==> currVel + (acc * t)
pos = currPos + velWithT ===> currPos + (velWithT * 1) [😱a t = 1 is still being applied to position calculation] 
```
By multiplying the original F with t, you're changing the acceleration calculation in applyForce to be a velocity calculation, this fixes the nextVelocity calculation (because its now adding a `acc * t` instead of `acc * 1`), but it doesn't address the final position calculation which is still using `velocity * 1`.

All this to say that, I think you found a legit bug in the calculations, awesome find!  But I think the solution to fix all the things would be to update `applyForce` to accept `force`, `entity`, and `duration`.  And then update the `nextVelocity` and `nextPosition` calculations to use the passed in duration (i.e. v = v0 + at, p = p0 + vt).

And then all you'd have to do is pass in duration = 0.001 to any call to `applyForce` and boom, that should fix the issue of calling each step on a millisecond basis but applying the acceleration & velocity at per second basis.

parkerziegler: Thanks so much for typing up your thought process @mariano-formidable ❤️ I think you hit the nail on the head, and this would explain some of the challenges I've run into moving this same setup over to friction. I'm going to make the change just for fluid resistance in this PR, then try to get it ported over for the other two in the next PR (which will also introduce the `infinite` flag).

src/animation/fluidResistance.ts
parkerziegler: * Scratch that, since it is in the `applyForce` function it'll need to come up separately, since I don't want to disrupt the other local changes I have here. I'm going to merge as is, and then make a dedicated PR to fixing this across all three forces. Thanks again for your help here!

PR #19, opened 2020-01-08
Fix timescale of force application.

parkerziegler:
This PR fixes a long standing bug I sense had been lingering and that @mariano-formidable proposed an awesome fix for! The key is realizing that our force calculations are performed in Newtons (N), which translate to kg * m / s^2. However, our `applyForceForStep` function runs once every ms, meaning that we were applying a second's worth of force every ms.

To correct this, we needed to alter our `applyForce` function to account for time. Using the following equations:

```
a = F / m;
v = v_current + at;
p = p_current + vt;
```

We see that we need to include time as a parameter to our `applyForce` equation and that, for most cases, we'll want it to be set to 0.001 (1 millisecond) 🎉 

In the case of `useGravity2D`, I decided to expose a new `timescale` parameter to allow a user to adjust this. If we use the actual 0.001 second timescale, the animations lose a bit of their luster (even though they are more accurate). This is because the gravitational force, compared to other forces we're modeling, is actually rather weak. A default of 0.1 (essentially, we apply 100x the force each ms) gives a bit more satisfying // expected results, at the expense of being mathematically scientific. I think this is a fine trade-off – users who want the _exact_ force application can use 0.001 as their timescale (and watch their orbs move very, very slowly 😂).


mariano-formidable:
Nice update!  And yes, real physics can often lead to less exciting results which is why in most game forces are hugely exaggerated.  Forces that would realistically generate 10N would be multiplied to generate 100N or more, just so it "feels better" and "plays better" :D 

I think your solution to allow for custom time for `gravity2D` makes sense, but I wonder if instead of having `timescale` replace the default ms time frame used, it could instead be a multiplier.  This is how game engines like Unity work (and they literally call it timeScale).  

If the timeScale is set to 1, then all physics operate as they normally would, but if you wanted to have a slow-mo type effect, you could set the timeScale to 0.5 and suddenly everything is moving at half speed.  If you set the timeScale to 0, you essentially are pausing the simulation, and, of course, if you set the timeScale to 10 or 100, you're speeding up the simulation by 10x to 100x  (like in Simcity type games where you can speed up time to watch your city grow quicker).

I think using timeScale like this (as a range from 0 - real bigger number range) for all the forces in renature would be a cool and useful feature.  It'll let users essentially speed up or slow down (or even pause) their animations by just updating the timeScale (they don't need to mess with mass or initial velocity or anything) which could add extra flexibility to make truly fun and creative animations/interactions.

Again, this is just a suggestion.  I think what you got and your fix is great as is, just food for thought for the future of the api.

parkerziegler:
I love the idea, and I think that's a better use for `timescale` – rather than a value it's a positive integer multiplier. I'll implement that now for `gravity2D`, and then re-approach it when we get to pausing animations / modifying the play state more specifically.

PR #20, opened 2020-01-09
Add formal support for an infinite parameter.

parkerziegler:
Alright, this PR is an interesting one, and I'm not positive this is the best approach, so I'd love thoughts and feedback on things here.

The core idea behind this PR is introducing an `infinite` config parameter to our animations. The core idea is that, rather than calling `stop` when we've reached the physics stopping condition, we actually reset the simulation and run it in the _reverse_ direction. So if we just slid a mover on a rough surface from 0m to 200m, `infinite` will immediately reset the animation and slide the mover from 200m to 0m using the same initial parameters. Same animation, different starting points.

With this change, I did introduce the notion of `playState`. `playState`, right now, just determines whether an animation is running in the forward or reverse direction, but I think it could also be quite useful for adding `Paused` support later on.

One of the trickiest bits with all of this was fluid resistance. Our simulation involves a dropping an object into a fluid, and calculating the net force applied by gravity and the drag force to determine position over time. When reversing this, we actually have to _invert_ gravity in our coordinate system (so it pushes in the up, or negative Y, direction). Not too tricky, just something to be aware of. I also haven't been able to yet arrive at a clean way of making `settle` work with `infinite` with fluid resistance because we don't have great criteria for saying, "Ah yes, the mover object has settled." I'm thinking a threshold velocity of something like 0.01 m/s could be "good enough" to say that the mover has come to rest and we can reverse the animation direction.

**Edit**

After a bit of digging, I think I found a good way to make `infinite` and `settle` play nicely. Basically, we just introduce an `isSettled` computation to check that the velocity of the `mover` has been reduced to <1 m/s when the `PlayState` and `position` have reached the reversal condition. This allows nice springy bounces before we start playing the animation in the reverse direction.


__tests__/forces/Gravity_test.re
```diff
@@ -15,9 +15,9 @@ describe("Gravity", () => {
 
     it(
       "should correctly derive the force of gravity at the Earth's surface", () => {
-      let massEarth = 5.97 *. Math.pow(~base=10, ~exp=24);
+      let massEarth = 5.97 *. 10. ** 24.;
```
parkerziegler: 🤦‍♂ OCaml has an exponentation operation already, `**`, so no need to have our own 🎉

src/animation/fluidResistance.ts
```diff
@@ -41,9 +47,45 @@ const applyFluidResistanceForceForStep = (
   });
 
   const gravitationalForce: Vector<number> = [0, mover.mass * gE];
-  const netForce = addf({ v1: dragForce, v2: gravitationalForce });
+  const netForce = addf({
+    v1: dragForce,
+    v2:
+      playState === PlayState.Forward
+        ? gravitationalForce
+        : multf({ v: gravitationalForce, s: -1 }),
```
parkerziegler: If running the animation in the reverse direction, we need to reverse the direction of the gravitational force. Drag will always operate in the reverse direction of velocity, so that's already taken care of ✅

src/animation/gravity.ts
```diff
@@ -36,31 +49,62 @@ const applyGravitationalForceForStep = (
...
+    state.mover = {
+      ...state.mover,
+      acceleration: [0, 0],
+      velocity: [0, 0],
+      position: [config.r, 0],
+    };
+    state.attractor = {
+      ...state.attractor,
+      position: [0, 0],
+    };
```
parkerziegler: In the case of gravity, when the animation completes we essentially transport the `attractor` to [0, 0] so the `mover` will reverse course and get pulled towards it.

PR #22, opened 2020-01-14
Add rollup to build library.

parkerziegler:
This PR adds [rollup](https://github.com/rollup/rollup) to build `renature` for distribution on npm 🎉 I mostly just copied my configuration from [`next-urql` ](https://github.com/FormidableLabs/next-urql/blob/master/rollup.config.js) since that's worked pretty well for me. That version is based off of `urql`'s `rollup` config minus all the `buble` stuff. When we really want to shave more bytes we can bring in `buble`.

Any feedback is welcome! I haven't yet tested this, but will locally by symlinking a little sandbox app. Once we get that verified and get Travis in, I think we can release v0.1.0-rc.0. I'm going to use rc's on this project as a way of getting features out quick without needing to contemplate semver challenges for every tiny thing.

@kitten and @JoviDeCroock tagging you b/c you seem to have a knack for shaving bytes 😉


rollup.config.js
```diff
@@ -0,0 +1,113 @@
...
+      ignoreGlobal: true,
+      include: /\/node_modules\//,
+      namedExports: {
+        react: Object.keys(require('react')),
+      },
+    }),
+    typescript({
+      typescript: require('typescript'),
+    }),
+    babel({ extensions, include: ['src/**/*'], exclude: 'node_modules/**' }),
```
JoviDeCroock: The output that will come out is ES6 and won't be compatible some browsers (IE11 edge,...)  this because tsconfig has ES2015 target and babel config has current node

parkerziegler: Good catch! I didn't even notice that, but in looking at the non-minified build output now there definitely is some ES6 in there 😅 🆒 I'll give `microbundle` a shot!

kitten: I think it’s also worth just copying what we have for urql and combining it with what we have in Wonka (it already looks pretty similar)

Happy to send some commits over for that!

Basically, we use TypeScript to compile to certain helpers but not much, we then use Buble because it handles classes and spreading/rest operators the best, lastly we use Babel only for the “detail” work, like making sure that we have a unified helper for `Object.assign`

We also recently added `closure-compiler` but that can mangle code which may not be a good choice for renature

parkerziegler: I'd love to have you send over some commits if you have the time! I did manage to get `microbundle` working with pretty minimal effort, but I'd like to see what you have in mine. Agreed that I don't think it's quite worth getting full closure compiler stuff in just yet, but everything else sounds good 😄

PR #38, opened 2020-02-08
Attempt to fix Travis build matrix for Node.

### LINKED ISSUE #36: CI: Multiple test jobs
We've got 3 Test jobs -- 2x node10, 1x node12 which seems weird. E.g. https://travis-ci.com/FormidableLabs/renature/builds/147818433

Screen Shot 2020-02-06 at 12 57 57 PM

parkerziegler:
Fix #36 

This is a first attempt at fixing the build matrix issues we were seeing with Travis where the Test stage was getting three jobs run (2x on Node 10, 1 on Node 12) and all other stages were only getting a single Node 10 build 🤔 Everything I read here: https://docs.travis-ci.com/user/build-matrix/ suggested that this setup should have worked fine, and yet...

I also realized we weren't building our Reason tests in the Test stage, so they were never getting run 😬 This fixes that by adding `yarn build:re` as the first step in the Test stage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,30 +22,24 @@ install:
 
 jobs:
   include:
-    - &buildre
-      stage: Build Reason
+    - &buildReTest
+      stage: Build Reason and Test
       script:
         - yarn build:re
-    - <<: *buildre
+        - yarn test
+    - <<: *buildReTest
       node_js: 12
-    - &buildts
-      stage: Build TypeScript
+    - &checkTs
+      stage: Check TypeScript
       script:
         - yarn check:ts
-    - <<: *buildts
+    - <<: *checkTs
       node_js: 12
     - &lint
       stage: Lint
       script:
         - yarn lint
     - <<: *lint
-      node_js: 12
-    - &testall
-      stage: Test All
-      script:
-        - yarn build:re
-        - yarn test
-    - <<: *testall
       node_js: 12
     - &build
       stage: Build
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,12 +40,12 @@ jobs:
         - yarn lint
     - <<: *lint
       node_js: 12
-    - &testAll
+    - &testall
       stage: Test All
       script:
         - yarn build:re
         - yarn test
-    - <<: *testAll
+    - <<: *testall
       node_js: 12
     - &build
       stage: Build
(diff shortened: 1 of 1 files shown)


ryan-roemer:
Not looking quite right in how the build is going. I'll take a stab this weekend or next week with maybe some reading and changes here to see if we can get what we want...

parkerziegler:
Ah yep, I think I figured out a bit about where the problem lies:

> Each job included in jobs.include inherits the first value of the array that defines a matrix dimension. In the example above, without explicitly setting rvm: 2.4, the included job inherits rvm: 2.3.

From https://docs.travis-ci.com/user/build-stages/#build-stages-and-build-matrix-expansion.

Basically, I don't think matrix expansion is really supported when defining jobs with `jobs.include` 😭 So I think we may have to do it manually like so:

```yml
- stage: Test Node 10
      node_js: 10
      script:
        - yarn build:re
        - yarn test
- stage: Test Node 12
    node_js: 12
      script:
        - yarn build:re
        - yarn test
```

Moreover, the three test jobs for test comes from, I think, the fact that Travis's default job is to run any script that looks "test-like" (in our case I think b/c we name it `Test`) automatically, so we get Travis's auto behavior for Node 10 and Node 12 on test, and then our actual Test job (with the `yarn build:re` step first) runs. :oof: this is pretty frustrating behavior and really could be more clearly documented.

parkerziegler:
Ok @ryan-roemer after a few hiccups I think it's all good!

image

I learned a bit about YAML anchors and aliases here: https://boneskull.com/mocha-and-travis-ci-build-stages/ and by looking at how Mocha does it: https://github.com/mochajs/mocha/blob/master/.travis.yml Pretty neat! I'm pretty happy with this solution overall if you are, feels like a nice compact way to do things given that Travis doesn't inherently expand the build matrix with `jobs.include`.

.travis.yml
```diff
@@ -10,35 +10,51 @@ branches:
...
+      stage: Lint
       script:
         - yarn lint
-    - stage: Test
+    - <<: *lint
+      node_js: 12
+    - &testall
+      stage: Test All
       script:
+        - yarn build:re
```
ryan-roemer: Because we have to build reason here for tests, it seems inefficient to also do a `buildre` stage that is essentially thrown away because we don't share artifacts. Maybe change this to `"Build reason and test"` or something?

Issue #40, opened 2020-02-23
Normalize interpolator functions.


parkerziegler:
As our interpolators have grown there's a bit of drift occurring in how they're implemented and the sorts of data structures they're returning, which makes them harder to reason about. To address this, we'll want to settle on a normalized structure for our interpolators focused specifically on the parsing and validation steps. For this, I think we can take inspiration from Popmotion's [`style-value-types`](https://github.com/Popmotion/popmotion/tree/master/packages/style-value-types).

In `style-value-types`, values can be tested using a `.test` method which returns a `boolean` indicating whether the supplied string is of that particular type. Each type also have a parser to parse the raw string into its component pieces, which is what we do in our `parse_` functions today. Finally, there's a `transform` function to convert a parsed string back to a raw string. I really like this structure and I think we ought to mimic it. I'll put up a PR showing how I envision this happening for our latest animating property, `box-shadow`, as a reference.

parkerziegler:
Closed by #41.

PR #41, opened 2020-03-01
Add support for interpolating CSS box-shadow.

parkerziegler:
**All reviews optional, just tagging folks who might be interested.**

This PR is an exciting one – we can now animate `box-shadow` in our hooks! This looks pretty dang neat.

200225_interpolate_box_shadow

The logic to support this is somewhat more complex than previous interpolations because `box-shadow` has a more complex formal syntax. The `box-shadow` page on MDN has a great formal spec, but it boils down to:

`<shadow> = inset? && <length>{2,4} && <color>?`

We use this to validate user-supplied `box-shadow`s before attempting to animate them.

This PR also beings to address #40. I'm taking a cue from Popmotion's [`value-style-types`](https://github.com/Popmotion/popmotion/tree/master/packages/style-value-types), which expose specific `test` and `parse` functions to validate user-supplied values and turn them into normalized structures. This works beautifully with Reason, since we can use records to explicitly defined normalized types for certain CSS properties. I'm going to backport this logic to other interpolators soon, such that each module has the same three set of functions.

- `test` – validate the string is valid for the CSS property being animated.
- `parse` – parse the string into a normalized format for that module.
- `remap` – interpolate the normalized structure from one value to another based on the physics `domain`.


mariano-formidable:
I scrubbed the code (in github) as best as I could, it looks good to me!  I pulled locally and saw the magic running in storybook, very awesome feature addition and nice parsing of all those box shadow properties!!! 👍

PR #42, opened 2020-03-13
Normalize parser and interpolator structure.

parkerziegler:
This PR adds support for a normalized `parser` and `interpolator` format. Here's the gist:

1. We added a new folder called `parsers`. Each Reason module in this folder is composed of between two and four functions. One (or two in the case of `box-shadow` and `transform`) of these functions is a `test` function – it determines if the string passed by the user matches the shape of a particular CSS property, and, to some extent, if its valid for that type. Is this shape a `box-shadow`, `transform`, a value with a unit, a color, etc? `test` will decide! The second (or third and fourth) function is a `parse` function. `parse` will, you guessed it, parse the string into a normalized format. For colors, `transform`s and `box-shadow`s we use Reason records to decompose these strings into their component parts.

2. The `interpolate` folder is now `interpolators`. The structure closely mirrors `parsers`. Each Reason module here has one (or two) functions used to interpolate CSS strings from one value to another. This was happening before, mostly just a more consistent naming structure.

Otherwise, changes are mostly related to adding tests, renaming things, and adding support for animating `box-shadow` 🌶 
renature_box_shadow


src/interpolaters/Interpolate_transform.re
```diff
@@ -0,0 +1,42 @@
...
+  let fromTransform = Parse_transform.parseTransform(dl);
+  let toTransform = Parse_transform.parseTransform(dh);
+
+  let transforms =
+    switch (
+      fromTransform.transform->Js.Nullable.toOption,
+      toTransform.transform->Js.Nullable.toOption,
+    ) {
+    | (Some(tl), Some(th)) =>
+      let tl_1 = Js.String.split(", ", tl);
```
kitten: Careful here 😅 the white space is optional can be repeated and you could have new lines. Time for another regex?

src/parsers/Parse_box_shadow.re
```diff
@@ -0,0 +1,123 @@
...
+      |> Js.Array.every(p =>
+           p->Belt.Option.getWithDefault("") |> Parse_unit.testUnit
+         )
+    | _ => false
+    };
+  };
+};
+
+let testBoxShadows = val_ => {
+  let boxShadows = Js.String.splitByRe(outerCommaRe, val_);
```
kitten: Seems like you’ll need this for transforms then?

src/parsers/Parse_box_shadow.re
```diff
@@ -0,0 +1,123 @@
...
+      blur: filteredProperties[2],
+      spread: "0px",
+      color: color^,
+      inset: inset^,
+    }
+  | 4 => {
+      offsetX: filteredProperties[0],
+      offsetY: filteredProperties[1],
+      blur: filteredProperties[2],
+      spread: filteredProperties[3],
```
kitten: Nit: you’ll get a “safe” get here any way which throws, so you can maybe ditch the pattern match here for a simple `Belt.Array.get` + `Belt.Option.getWithDefault`

src/interpolaters/Interpolate_transform.re
parkerziegler: Right, that's a good shout, so we need to support `translate(10px,10px)` as well. Can you give a newline example? Do you mean we could get something like:

```js
from: {
   transform: `translate(20px,
     20px)`
}

I could see just splitting by the comma and then trimming the whitespace in lieu of another regex here, but curious to hear your thoughts.

src/parsers/Parse_box_shadow.re
parkerziegler: Good shout, I like that ✅

src/parsers/Parse_box_shadow.re
parkerziegler: I have a specific regex for transforms, but it's quite similar! It's down in `Parse_transform`. The goal for this regex is to split on commas not inside parenthesis. So it handles when a user supplies multiple box-shadows that use `rgba` or `hsla` colors (colors don't get split, but the comma separating the individual box-shadows gets picked up).

src/interpolaters/Interpolate_transform.re
parkerziegler: The no whitespace case should be addressed by this commit: https://github.com/FormidableLabs/renature/pull/42/commits/a88260ca4c00a783d40a834d613c1bdb3bb488f6

PR #46, opened 2020-03-30
Migrate to react-static v7.

parkerziegler:
This PR is enormous and unreviewable. For that I'm sorry 😞 For anyone curious, this PR:

- Upgrades us from `react-static` v5 to v7.
- Revises some non-semantic CSS we had powering our landing page.
- Integrates [`react-static-plugin-md-pages`](https://github.com/kitten/react-static-plugin-md-pages) and updates our docs structure to fit with this new architecture.
- Rewrites a good portion of the docs and integrates MDX with live code previews using `react-live` 💅 

Lots of good stuff! I was fairly aggressive with the changes, mostly porting over all of the goodness from the `urql` docs. We may find trailing unneeded bits of code and assets here, but for the most part this should make for a much nicer docs experience. Some screenshots!

image

image

image

Once the staging build goes up I'd love to get Joe or another member of the design team to do a design pass.


docs/src/components/navigation.js
```diff
@@ -1,94 +1,137 @@
...
+  margin: ${p => p.theme.spacing.xs} 0;
+  color: ${p => p.theme.colors.text};
+  font-weight: ${p => p.theme.fontWeights.heading};
+  text-decoration: none;
+  width: 100%;
   &:hover {
-    color: ${({ theme }) => theme.colors.linkLightHover};
+    color: ${p => p.theme.colors.accent};
+  }
+  &.active {
```
kale-stew, line 73: I think you might have a typo here, did you mean `&:active`?

docs/src/components/section-title.js
```diff
@@ -6,11 +6,11 @@ export const SectionTitle = styled.h2`
   font-size: 2.5rem;
   flex: auto;
   line-height: 1.3;
-  margin: 2rem 0 3rem;
+  margin: 4 rem 0;
```
kale-stew, line 9: I think you might have an extra space here 🙂

docs/src/screens/home/features.js
```diff
@@ -6,41 +6,88 @@ import { BodyCopy } from '../../components/body-copy';
...
   }
-  @media (min-width: 1024px) {
-    width: calc(1 / 3 * 100% - (1 - 1 / 3) * 80px);
+
+  @media ${p => p.theme.media.sm} {
+    max-width: none;
+    flex-direction: row;
+
+    > * {
+      margin-top: 0;
```
kale-stew, line 33: Is there an advantage to explicitly setting every margin property vs `margin: 0`? Does this allow for more explicit override?

parkerziegler: Good shout! Yep you got it, the main reason is to allow your child elements to maintain any sort of left or right margin that they may be getting from another class. With this stack layout you really want to enforce vertical spacing without affecting the horizontal. More useful when it comes to an abstracted mixin in Sass – doesn't matter as much here when all classes this selector is being applied to are already in the same file.

docs/src/components/section-title.js
parkerziegler: Wowza, awesome catch, thank you!

parkerziegler:
@boygirl I merged away after verifying the staging build 😄 I think things are mostly in good shape!

PR #49, opened 2020-04-15
Change import of useFriction to useGravity in useGravity example

wcrozier12:
This PR changes the import for the useGravity example from useFriction to useGravity


parkerziegler:
Ah good catch, thanks man! I wanted to have the `import` statements in there so folks could copy and paste right to CodeSandbox, but they actually get stripped before being sent to `react-live` since `react-live` can't interpret ES6 `import`. Thanks a bunch!

PR #50, opened 2020-04-17
Add G parameter to useGravity and useGravity2D hooks.

parkerziegler:
This PR adds `G` as a parameter for the `useGravity` and `useGravity2D` hooks. Previously we only used the actual Universal Gravitational Constant, G, to derive the magnitude of the force of gravity when running our physics simulations. However, since real G is so small, users would have to supply mass parameters on the order of magnitude of planets. This is great for realistic simulations, but not necessarily for readability or accessibility when users are just playing around.

Now, users can supply their own value of `G` to use if they prefer. We'll continue to default to real G for precision and sticking with the spirit of keeping `renature` mathematically accurate, but I do think this may help to reduce confusion or ease of use around this hook for a lot of users.

While I was here I also upgraded some dependencies, including the BuckleScript compiler to pick up on some of the nice optimzations recently introduced. **You can ignore any files in this ending with .bs.js or .gen.tsx**.


__tests__/forces/Gravity_test.re
```diff
@@ -7,7 +7,12 @@ describe("Gravity", () => {
     it("should derive the force of gravity between two objects", () =>
       Expect.(
         expect(
-          Gravity.gravityForceMag(~attractorMass=1., ~moverMass=1., ~r=1.),
+          Gravity.gravityForceMag(
+            ~attractorMass=1.,
+            ~moverMass=1.,
+            ~r=1.,
+            (),
```
parkerziegler, line 14: `unit` parameter (the last `()`) has to be added to this function now that it takes an optional argument. This is how we signal that the function has been fully applied even tho we didn't supply an argument for `G`, since Reason will automatically curry functions.

__tests__/forces/Gravity_test.re
```diff
@@ -74,5 +95,77 @@ describe("Gravity", () => {
...
+            ~g=1.,
+            (),
+          )
+          |> snd,
+        )
+        |> toBeCloseTo(-0.6413136576)
+      )
+    );
+
+    it(
```
parkerziegler, line 133: Adding additional tests for untested functionality. For those curious, the `threshold` parameter limits the magnitude of the gravitational force by supplying a range within the force is applied. If the `mover` exceeds the upper bound of the `threshold` parameter (i.e. upper bound is 40 and the `mover` is 50m away from the `attractor`), we supply a force equal to the `mover` being at 40m away. This is how we ensure the `mover` never goes flying off the screen. If this parameter isn't specified you can get behavior where the `mover` slingshots off quite a ways 😂

stories/useGravity2D.stories.tsx
```diff
@@ -48,7 +48,52 @@ export const Gravity2DBasic: React.FC = () => {
...
+    const root = document.getElementById('root');
+    if (root) {
+      setCenter([root.clientWidth / 2, root.clientHeight / 2]);
+    }
+  }, []);
+
+  const [props] = useGravity2D({
+    config: {
+      attractorMass: number('attractorMass', 20),
+      moverMass: number('moverMass', 1),
```
parkerziegler, line 64: With a `G` of 0.4 our masses can now be a lot smaller to achieve a similar effect animation-wise.

PR #60, opened 2020-05-24
Add proposal for onFrame and onAnimationComplete APIs.

parkerziegler:
This PR adds in two new ways of interacting with the `renature` API – `onFrame` and `onAnimationComplete`. This is one way I've thought of to support a few use cases in a single API:

- Allow users to execute their own logic in response to a new frame or animation completion.
- Hand users a `progress` value between 0 and 1 that represents the state of the physics tween.

The API looks like this:

```typescript
useFriction<HTMLDivElement>({
  ...,
  onFrame: progress => {
    /* Execute some logic in response to a frame update. */
  },
  onAnimationComplete: () => {
    /* Execute some logic now that the animation has finished. */
  },
});
```

`onFrame` currently receives `progress` as a value. This is the only API I've thought of where we can safely access `progress` as a `ref` value tracked in the animation functions themselves. We can't directly return `progress` from the hook itself, because we'd then need some way for the calling component to re-render in order for new values of `progress` to get picked up. And of course, with declarative animations like ours they'll restart if the component re-renders.

@NgoKnows take a look at the new example I added for accessing `progress` in `onFrame` and let me know if you think this would suit your needs! I'm not confident that this is necessarily the best API – just that it can solve certain problems by shifting more flexibility to the user. @littlemilkstudio would also appreciate your thoughts here!

200524_progress

### DIFF
--- a/src/hooks/useFriction.ts
+++ b/src/hooks/useFriction.ts
@@ -66,7 +66,7 @@ export const useFriction = <M extends HTMLElement | SVGElement = any>({
           }
 
           if (onAnimationComplete) {
-            onAnimationComplete(1);
+            onAnimationComplete();
           }
         });
       },
--- a/src/hooks/useFriction.ts
+++ b/src/hooks/useFriction.ts
@@ -14,6 +14,8 @@ export const useFriction = <M extends HTMLElement | SVGElement = any>({
   immediate = true,
   delay,
   infinite,
+  onFrame,
+  onAnimationComplete,
 }: UseFrictionArgs): [
   { ref: React.MutableRefObject<M | null> },
   Controller
@@ -28,25 +30,28 @@ export const useFriction = <M extends HTMLElement | SVGElement = any>({
 
   const { controller } = React.useMemo(() => {
     const interpolators = getInterpolatorsForPairs({ from, to });
+    const maxPosition = getMaxDistanceFriction({
+      mu: config.mu,
+      initialVelocity: config.initialVelocity,
+    });
 
     return friction1D({
       config,
       onUpdate: ({ position }) => {
         interpolators.forEach(({ interpolator, property, values }) => {
           const value = interpolator({
-            range: [
-              0,
-              getMaxDistanceFriction({
-                mu: config.mu,
-                initialVelocity: config.initialVelocity,
-              }),
-            ],
+            range: [0, maxPosition],
             domain: [values.from, values.to],
             value: position[0],
           });
 
           if (ref.current) {
             ref.current.style[property as any] = `${value}`;
+          }
+
+          if (onFrame) {
+            const progress = position[0] / maxPosition;
+            onFrame(progress);
           }
         });
       },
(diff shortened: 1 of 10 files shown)


src/hooks/useFriction.ts
```diff
@@ -28,26 +30,29 @@ export const useFriction = <M extends HTMLElement | SVGElement = any>({
 
   const { controller } = React.useMemo(() => {
     const interpolators = getInterpolatorsForPairs({ from, to });
+    const maxPosition = getMaxDistanceFriction({
+      mu: config.mu,
+      initialVelocity: config.initialVelocity,
+    });
```
parkerziegler, line 36: We should see a slight performance boost moving this calculation out of the frame loop 😅

src/hooks/useFriction.ts
```diff
@@ -28,26 +30,29 @@ export const useFriction = <M extends HTMLElement | SVGElement = any>({
...
           });
 
           if (ref.current) {
             ref.current.style[property as any] = `${value}`;
           }
+
+          if (onFrame) {
+            const progress = position[0] / maxPosition;
+            onFrame(progress);
+          }
```
parkerziegler, line 55: Here's where we'll execute the `onFrame` logic if provided. This feels a bit cleaner than returning `progress` as a third part of the tuple returned by this hook. I'm also not sure we could really pull that off. The parent component would have to re-render to see the updated value of a progress `ref`, which would restart the animation 😬

stories/useFriction.stories.tsx
```diff
@@ -267,3 +263,28 @@ export const FrictionSVG: React.FC = () => {
...
+    config: {
+      mu: number('mu', 0.1),
+      mass: number('mass', 30),
+      initialVelocity: number('velocity', 10),
+    },
+    onFrame: progress => {
+      if (progressRef.current) {
+        progressRef.current.innerHTML = `${progress.toFixed(2)}`;
+      }
+    },
```
parkerziegler, line 286: @NgoKnows here's how we pull that demo off – essentially, attach a ref to the `span` so we can edit the `innerHTML`. We can't store this in a `state` value, because then we'd be re-rendering this component on every frame.

littlemilkstudio:
```
useFriction<HTMLDivElement>({
  ...,
  onFrame: progress => {
    /* Execute some logic in response to a frame update. */
  },
  onAnimationComplete: () => {
    /* Execute some logic now that the animation has finished. */
  },
});
```

Hey @parkerziegler, yea this looks great! 

`onAnimationComplete` looks exactly like the native `animationend`: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/animationend_event

And the setup above for it for it allows a component that uses renature under the hood, to have an external facing api very similar to a css animated component which is always nice.

And yea either providing this or returning a subscription method are the 2 options I can think of. The subscription approach is nice if you need access to this callback in other places in the component tree other than the call site (if you wanted multiple components to be aware of it's progress).

That being said, this setup can already utilize a subscription, it's just unopinionated on the users subscription system of choice:

```
const emitter = useMemo(() => {
  ...
  return {
    emit,
    subscribe
  }
}, [])

useFriction<HTMLDivElement>({
  ...,
  onFrame: emitter.emit,
  onAnimationComplete: () => {
    /* Execute some logic now that the animation has finished. */
  },
});

useEffect(() => {
  const unsubscribe = emitter.subscribe(progress => {
    /* do something with progress */
  });
  return unsubscribe;
}, [emitter.subscribe])
```

The user can create `emitter` however they'd like. Can use whatever package they fancy or utilize the native api:
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget

PR #68, opened 2020-06-14
Fix #15. Add GPU compositing for transform animations by default.

### LINKED ISSUE #15: GPU Compositing
Research how to force the compositing to the GPU instead of the CPU.

parkerziegler:
Fix #15 

This PR adds default support for GPU compositing of `transform` animations. When a user animates a node using `transform`, `renature` will automatically append `translateZ(0)` to the animation unless:

- The user expressly disables this behavior using `disableHardwareAcceleration`.
- The user has already provided `translateZ` as an animating property.

As we can see now, animations specifying `transforms` get put on their own layer so they can be uploaded to the GPU (all the other layers there are Storybook extras):

image

I know that `opacity` animations theoretically can also be uploaded to the GPU, but not sure how to achieve that (or if it's already happening). In general, would love feedback and guidance on:

1) Whether or not this is a good strategy for GPU compositing. I've read here and there and seen different advice on GPU compositing. This approach is mostly inspired by `framer-motion` and some reading in the `react-pose` codebase.
2) Ways to verify GPU compositing and analyze performance.


src/parsers/pairs.ts
```diff
@@ -104,8 +110,17 @@ export function getInterpolatorForPair({ from, to }: CSSPairs) {
         interpolator: interpolateTransforms,
         property: fromProperty,
         values: {
-          from: fromValue,
-          to: toValue,
+          from:
+            // Add translateZ(0) to transform if it's not already present
+            // and if hardware acceleration is enabled.
+            fromValue.indexOf('translateZ') === -1 &&
```
parkerziegler, line 116: Decided to use `indexOf` in lieu of `includes` for IE11 compatibility. Not sure if folks have opinions here (i.e. consumer should polyfill `String.prototype.includes`, we just shouldn't care about IE11, etc)?

PR #72, opened 2020-06-30
use formideploy for deploy:prod

keithcom:
initial change to formideploy for production plus cleanup of old deploy scripts


ryan-roemer (approved):
Looks good!

OK, once we merge, let's watch https://tmp-experiment-02.formidable.com/open-source/renature/ and verify it ends up with content (currently a 404).

PR #73, opened 2020-07-01
add github deployment token

keithcom:
add token


ryan-roemer (approved):
We're in business! 

Screen Shot 2020-07-01 at 10 40 38 AM

After merge to our WIP we should see both a staging message and production message because of our testing setup.

PR #74, opened 2020-07-04
Add use<Force>Group hooks.

parkerziegler:
This PR is a really exciting one! It adds support for `use<Force>Group` hooks (i.e. `useFrictionGroup`, `useGravityGroup`, etc.) The API looks like this:

```ts
const [nodes, controller] = useFrictionGroup(5, i => ({
  from: {
    transform: 'translateY(0px)',
    background: 'teal',
    borderRadius: `${Math.floor(Math.random() * 100)}%`,
  },
  to: {
    transform: 'translateY(100px)',
    background: 'orange',
    borderRadius: `${Math.floor(Math.random() * 100)}%`,
  },
  config: {
    mu: 0.5,
    mass: 300,
    initialVelocity: 10,
  },
  delay: i * 500,
  infinite: true,
}));
```

Users can now, in a fine-tuned way, control multiple animating elements at once from a single hook! What's great is that this implementation can replace our basic `useFriction` hook by making `useFriction` just call out to `useFrictionGroup` with n=1.

**Note: I only updated `useFriction` for this PR. Didn't want to update everything until we feel good about the approach. Once approved, I'll bring this same logic over to the other hooks.**

## API Changes

There aren't too many API changes here other than the introduction of `controller.pause`. `controller.pause` is very much like `controller.stop`, but it preserves animation state. Conversely, `controller.stop` will fully clear the animating element from the `Set` of animating elements, so that element cannot be animated again unless the hook fully re-renders. I think this is a nice delineation to have and ensures we free up resources in a more explicit way.

One last piece of the API that I'd like feedback on. With this new model, we have `pause` as a config option and `controller.pause` as a function. As a config option, `pause` instructs the animation **not to run on mount**, since animations execute on mount inside of `useLayoutEffect` by default. Do we think this is confusing? Should we use a term like `runOnMount` with a default of `true`? `framer-motion` does a similar thing, allowing you to specify `initial: false` to disable run on mount animations.

## Bug Fixes 🐛 

I was also able to fix a bug in the course of discovery here 😱 Previously, calling `controller.start` did not check if our `rAF` function was already running. This meant a user could keep calling `controller.start` and new `rAF` calls would get fired for each one, leading to n number of `requestAnimationFrame` calls running concurrently. We now check if there is an existing frameloop before calling `start` 😅


src/animation/friction.ts
```diff
@@ -66,79 +64,107 @@ function reversePlayState(
...
 }
 
+function checkFrictionStoppingCondition({
+  state,
+}: StatefulFrictionAnimatingElement) {
+  return state.mover.velocity[0] <= 0;
+}
+
+let isFrameloopActive = false;
+const animatingElements = new Set<StatefulFrictionAnimatingElement>();
```
parkerziegler, line 74: Here's the money marker right here, an idea taken from `react-spring`'s implementation: https://github.com/react-spring/react-spring/blob/master/src/animated/FrameLoop.ts We store all animating elements in a `Set`. When the function below is called, we iterate through all nodes handed to it and add them to the `Set`. `controller.stop` then performs the action of removing elements from the `Set` when they're no longer animated. Beautifully simple!

src/animation/friction.ts
```diff
@@ -66,79 +64,107 @@ function reversePlayState(
...
+  let startFn: () => void = () => {};
+  let pauseFn: () => void = () => {};
+  let stopFn: (element: StatefulFrictionAnimatingElement) => void = () => {};
+
+  // Only start the frameloop if there are elements to animate.
+  if (animatingElements.size > 0) {
+    const { start, stop } = rAF();
+
+    startFn = () => {
+      if (!isFrameloopActive) {
```
parkerziegler, line 127: Here's where we check if there is an already executing `rAF` setup. If so, no need to `start` the frameloop again, as each call to `update` will "see" new elements added to our `Set` of animating elements and ensure they start moving. Verified in Performance tabs on Chrome and FF that we only get one `rAF` call for n elements!

src/hooks/useFriction.ts
```diff
@@ -29,95 +17,17 @@ export const useFriction = <M extends HTMLElement | SVGElement = any>({
...
-        });
-
-        if (onAnimationComplete) {
-          onAnimationComplete();
-        }
-      },
-      infinite,
-    });
-  }, [
+  const [props, controller] = useFrictionGroup(1, () => ({
```
parkerziegler, line 20: 🎉 `useFriction` becomes much tinier!

src/rAF/update.ts
```diff
@@ -0,0 +1,82 @@
...
+  applyForceForStep: (animatingElement: A) => Entity;
+  checkStoppingCondition: (animatingElement: A) => boolean;
+}
+
+/**
+ * The core updater function ran inside requestAnimationFrame.
+ * Iterates through all animating elements in the Set and updates
+ * their local animation state.
+ */
+export function update<A extends StatefulAnimatingElement>({
```
parkerziegler, line 16: `update` is now a generic function that will handle all hooks. Previously, a lot of this logic was duplicated across all hooks, so I pulled it into a single function. Now, calling functions pass in callbacks to handle animation reversal, stopping conditions, and force application.

parkerziegler:
The build failure is expected for now, since the other hooks have not been updated. I'll add them to this PR after we review a bit!

PR #75, opened 2020-07-08
Fix transpilation of for of loops with Set caused by dangerousForOf in Buble config.

parkerziegler:
This PR fixes a major 🐛 in v0.6.0. Previously, we had [`dangerousForOf`](http://buble.surge.sh/guide/#dangerous-transforms) enabled in our Buble transforms. `dangerousForOf` assumes all structures being iterated with `for...of` loop are array-like objects with a length property 😱 . This doesn't work for us because we iterate over a `Set` using `for...of`.

This PR adds in the necessary Babel transform to support IE11: https://babeljs.io/docs/en/babel-plugin-transform-for-of For now I'd like to keep IE11 support, but it is clear from the example that we end up adding a pretty hefty transpilation cost with this plugin. In a future release of `renature` (likely v1), we'll remove IE11 support entirely and strip this plugin. But for now, it makes sense to keep it in.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
   "license": "MIT",
   "devDependencies": {
     "@babel/core": "^7.10.1",
-    "@babel/plugin-transform-for-of": "^7.10.4",
     "@babel/plugin-transform-modules-commonjs": "^7.10.1",
     "@babel/plugin-transform-object-assign": "^7.10.1",
     "@babel/preset-env": "^7.10.1",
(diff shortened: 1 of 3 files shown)


kitten:
In urql we'd typically not use `for...of` but would hand write for-each loops instead. Given that you're about to pull in a lot of helper code potentially for the iterables, is that maybe a preferable option?

parkerziegler:
Good shout. The more I think about it too, the more I'm less inclined to include bundle bloating code for IE11. All modern browsers natively support `for...of` so I actually think I just won't include any transforms for it ✅

Issue #86, opened 2020-08-05
Add controller.set API.


parkerziegler:
Other animation libraries like [`react-spring`](https://www.react-spring.io/docs/hooks/use-spring) have a `controller.set` API. `set` is used to imperatively animate to a new state. This is an alternative to re-rendering the component with changed props.

Currently in `renature`, we support re-rendering the component with changed props. However, re-rendering does revert the animation to the `from` state (which is a separate issue to tackle). We do want some way to imperatively set a component to a particular state, and animate smoothly from the current state to that new state.

### Proposal

I'm thinking the API would look something like the following:

```js
import { useFriction } from 'renature';

const Component = () => {
  const [props, controller] = useFriction({
    from: {
      transform: 'translateX(100px)'
    },
    to: {
      transform: 'translateX(200px)'
    }
  });

  return (
    <div
      {...props}
      onMouseEnter={() => controller.set({ transform: 'scale(1.05)' })}
      onMouseLeave={() => controller.set({ transform: 'scale(1)' }
    />
};
```

An animation like the above would:

- Animate when initially mounted, starting at the `from` and transitioning to the `to`.
- On `mouseenter`, would animate the `transform` while preserving the previously applied `translateX`.
- On `mouseleave`, would animate the `transform` while preserving the previously applied `translateX`.

`set` should always preserve animated values of unrelated properties, but overwrite any properties that are being mutated. For example, if we called `set({ transform: 'translateX(300px)' })` in the above, the `translateX(200px)` state defined in `from` should get overwritten.

PR #90, opened 2020-09-13
Add support for recovering from interrupted animations.

### LINKED ISSUE #89: Interrupted animations on state change.
Currently when an animation in `renature` is interrupted, it restarts the animation from the beginning. This is undesirable for the most common case of animation interruption, which is a state update meant to re-run the animation in the reverse direction (i.e. when a toggle is clicked). In these cases, we'd like animations to _not_ re-run from their `from` to their `to` property, but instead start from their current interrupted position and animate to their `to` property. An example from the docs site will make this clear:

renature-interrupted-animation

The goal for this type of animation should be to start from its _current_ state when a new animation is triggered rather than hard resetting to its `from` state. This will be important for better supporting reactive animations as well.

parkerziegler:
Fix #89 

This PR addresses issues with interrupted animations. Previously in `renature`, if some value in the `from`, `to`, or `config` properties of a hook changed, the animation would restart from the beginning. You can see a GIF of this in the linked issue. This was problematic for reactive animations, i.e. toggle or hover effects. In these cases, you want your animation to resume running from the animating element's last known set of animated styles rather than restarting from the beginning.

To fix this, I'm introducing the notion of an animated values cache. In this setup, we cache the interpolated animated styles of the simulated mover on each frame. This means that, if an animation gets interrupted, we start the new animation from the cached set of styles. An example – say we're moving between `opacity: 1` and `opacity: 0`. An animation gets interrupted and now it's instructed to animate instead to `opacity: 0.25`. At this moment, its current interpolated value (sat, `opacity: 0.5`) is cached. On the next re-render, rather than restarting the animation from `opacity: 1`, `renature` will start it at `opacity: 0.5`. This allows for animations that feel smooth and reactive to user input.

200913_renature_interrupted_animations

I'd be curious to get feedback on this approach. I like that it can help reactive animations feel a lot smoother, and it opens the way to address the `controller.set` API discussed in #86. The downside is that motion values themselves (`velocity` in particular) are not preserved across updates. This is a worthwhile tradeoff IMO. If we preserved `velocity`, there's a chance enough updates would result in atypical velocity values that would send a mover past it's possible range (which is calculated ahead of the animation running based on its configuration). This would result in animations that never reach their `to` state and then jump to it when `onComplete` is called – no good.

I am seeking active feedback here for any clever ideas! If the approach looks good I'll replicate to other hooks in this PR before merging.

### DIFF
--- a/src/hooks/useFrictionGroup.ts
+++ b/src/hooks/useFrictionGroup.ts
@@ -1,4 +1,11 @@
-import { createRef, useMemo, useLayoutEffect, useCallback } from 'react';
+import {
+  createRef,
+  useRef,
+  useMemo,
+  useLayoutEffect,
+  useCallback,
+  CSSProperties,
+} from 'react';
 
 import { CSSPairs, getInterpolatorsForPairs } from '../parsers';
 import {
@@ -12,6 +19,12 @@ import {
 } from '../animation';
 import { getMaxDistanceFriction } from '../forces';
 
+let _id = 0;
+const cache = new Map<
+  string,
+  Record<number, Record<keyof CSSProperties, any>>
+>();
+
 export type UseFrictionArgs = CSSPairs &
   HooksParams & {
     config?: FrictionConfig;
@@ -21,6 +34,8 @@ export const useFrictionGroup = <E extends HTMLElement | SVGElement = any>(
   n: number,
   fn: (index: number) => UseFrictionArgs
 ): [{ ref: React.RefObject<E | null> }[], Controller] => {
+  const hookId = useRef(`friction_${_id++}`);
+
   const { elements, start, stop, pause } = useMemo(() => {
     const animatingElements: AnimatingElement<
       FrictionConfig,
@@ -35,7 +50,7 @@ export const useFrictionGroup = <E extends HTMLElement | SVGElement = any>(
       // Derive interpolator functions for the supplied CSS properties.
       const interpolators = getInterpolatorsForPairs(
         {
-          from: props.from,
+          from: cache.get(hookId.current)?.[i] ?? props.from,
           to: props.to,
         },
         props.disableHardwareAcceleration
@@ -63,6 +78,18 @@ export const useFrictionGroup = <E extends HTMLElement | SVGElement = any>(
             const progress = position[0] / maxPosition;
             props.onFrame(progress);
           }
+
+          // Update the global cache of derived animation values.
+          const currentCacheValue =
+            cache.get(hookId.current)?.[i] ||
+            ({} as Record<keyof CSSProperties, any>);
+
+          cache.set(hookId.current, {
+            [i]: {
+              ...currentCacheValue,
+              [property]: value,
+            },
+          });
         });
       };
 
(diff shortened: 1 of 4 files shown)


maxyinger (approved):
Fun approach. lgtm!

src/hooks/useFrictionGroup.ts
```diff
@@ -35,7 +50,7 @@ export const useFrictionGroup = <E extends HTMLElement | SVGElement = any>(
       // Derive interpolator functions for the supplied CSS properties.
       const interpolators = getInterpolatorsForPairs(
         {
-          from: props.from,
+          from: cache.get(hookId.current)?.[i] ?? props.from,
```
hartzis: ❓ What if `n` changes? The cache lookup key could be wrong based on index?

💭 add `n` to the key somehow? and/or a `useEffect` that cleans/resets cache if `n` changes?

mariano-formidable (approved):
I think the concept is sound, but might get complicated to keep cache accurate as _any_ prop changes.  But it works as is right now 🎉

src/hooks/useFrictionGroup.ts
parkerziegler: Good question and good idea! Let me give you a full picture of the caching setup now.

So in this PR we create a global cache of animation values, which allows us to track these values for an arbitrary number of elements created by an arbitrary number of `useFriction` / `useFrictionGroup` hooks in your React tree. It basically looks like this:

```js
new Map([
  [
    // The first instantiated hook in the tree.
    'frictionGroup_0', {
      // Current animation values for the first element in the first hook in tree.
      0: {
        opacity: 0.50,
        transform: 'translateX(20px)'
      },
      // Current animation values for the second element in the first hook in tree.
      1: {
        opacity: 0.40,
        transform: 'translateX(15px)'
      }
  ],
  [
    'frictionGroup_1',
    {
      // Current animation values for the first element in the second hook in tree.
      0: {
        transform: 'scale(0.75)'
      }
    }
  ],
]);
```

Looking at it now, I think this cache is actually more information than we need. Each hook can be responsible for creating its own local cache of animation values that would get cleaned up when the component using it unmounts.

This then would allow us to directly use your idea of combining `n` and `i` into a cache key quite nicely! My only thought is that this would mean that adding / removing elements by altering `n` would force all animations to start from their `from` config (since they'd have no cached entry), which may or may not be the experience we want compared to adding / removing elements only affecting the added / removed element and allowing the others to continue animating fluidly. I'm going to run some experiments on `react-spring` and see how it works over there.

Great shout @hartzis, appreciate you taking the time to review!

parkerziegler: Following up here, `react-spring` appears to fully restart all animations whenever n changes: https://codesandbox.io/s/zealous-borg-2ykl8?file=/src/App.tsx, so that's at least a good starting place for us to think about what we want our behavior to be. It's likely a bit safer to do that, but I want to have a longer think (outside the scope of this PR though).

hartzis (approved):

Issue #108, opened 2021-01-24
Add eslint-plugin-import.


parkerziegler:
This is a simple one — rather than manually maintaining import rules, we should just add [`eslint-plugin-import`](https://github.com/benmosher/eslint-plugin-import) to handle this for us.

The most important rules for us are: 

- [`order`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md)
- [`newline-after-import`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/newline-after-import.md)

jzandag:
can I try doing this?

parkerziegler:
Hey @jzandag, sure thing. Here are the steps I think we need to take:

1. Add `eslint-plugin-import` to `devDependencies`.
2. Add the default `extends` from that plugin, in addition to its TypeScript support. This will amount to adding these three lines in the `extends` array in our root `.eslintrc.js`:

```js
'plugin:import/errors',
'plugin:import/warnings',
'plugin:import/typescript',
```

These should go after `'plugin:react/recommended'` and before `'prettier'`.

3. Add the following lines to the `rules` array:

```js
'import/newline-after-import': 2,
'import/order': [
  'error',
  {
    'newlines-between': 'always',
  },
],
```

4. Run `yarn lint --fix` to fix any import errors we may get and commit the results along with the above changes!

Does that work for you?

jzandag:
Thanks for the detailed instructions! I'll try doing it now

parkerziegler:
Fixed by #110

PR #110, opened 2021-01-27
Added eslint-plugin-import to devDependencies, and added some setting…

jzandag:
…s to .eslintrc.js

Reference to Issue #108


src/parsers/pairs.ts
```diff
@@ -1,10 +1,5 @@
-import type { CSSProperties } from 'react';
+import type { CSSProperties, RefObject } from 'react';
```
parkerziegler: You can remove the `RefObject` import here, this is why the CI run is failing 😄 

```suggestion
import type { CSSProperties } from 'react';
```

parkerziegler:
Awesome, thanks for this @jzandag! Just one last `import` to fix to get the build green, then LGTM!

parkerziegler:
Thanks again @jzandag!

jzandag:
@parkerziegler No problem! Thanks too for the learning opportunity!

PR #131, opened 2021-03-22
Task: update storybook dep to v ^6

gingerrific:
This PR updates our Storybook to v6 and handles any best practice or breaking changes updates. Changes include:

* remove `addons.js` and `config.js`, `webpack.config.js` in favor of `main.js`
  * typescript is now handled by default out of the box, so webpack config was removed 
* updated `getRandomHex` to including `padStart` to prevent error when function generates a value of only 5 characters
* update all storybook deps to v^6
* remove unused addons, `links` and `actions` cc @parkerziegler not sure if I missed this being used somewhere, but verified storybook functionality for all existing stories


parkerziegler (approved):
Awesome work @gingerrific!

PR #132, opened 2021-03-24
feat: add API for accessible animations

parkerziegler:
This PR is an exciting one — it adds support for accessible animations! You can now specify a `reducedMotion` configuration and, if a user prefers reduced motion in their OS accessibility settings, they'll see the animation using the `from` and `to` specified in `reducedMotion`.

The implementation of `usePrefersReducedMotion` is lightly modified from [Josh Comeau's excellent blog post](https://www.joshwcomeau.com/react/prefers-reduced-motion/) on the subject. What I appreciate about Josh's approach is that it also ensures that folks using devices / OSes that don't support this setting will still get the reduced motion animation (if specified). In this way, we go for accessible animations as the _default_ behavior rather than the _opt in_ behavior.

To use this API, users of `renature` can specify something like the following:

```typescript
export const FrictionInfinite: FC = () => {
  const [props] = useFriction<HTMLDivElement>({
    from: {
      background: '#f25050',
      transform: 'scale(1) rotate(0deg)',
    },
    to: {
      background: '#a04ad9',
      transform: 'scale(1.5) rotate(720deg)',
    },
    reducedMotion: {
      from: {
        opacity: 0,
      },
      to: {
        opacity: 1,
      },
    },
    repeat: Infinity,
  });
  return <div className="mover mover--red" {...props} />;
};
```

Lastly, this implementation also sets up event listeners to listen for changes to the `prefers-reduced-motion` setting. This means that an animation will alter itself mid-execution of this preference changes!

20210323_reduced_motion

### DIFF
--- a/__tests__/hooks/usePrefersReducedMotion.spec.ts
+++ b/__tests__/hooks/usePrefersReducedMotion.spec.ts
@@ -1,0 +1,53 @@
+import { renderHook } from '@testing-library/react-hooks';
+
+import { usePrefersReducedMotion } from '../../src/hooks/usePrefersReducedMotion';
+
+describe('usePrefersReducedMotion', () => {
+  it.each([
+    { query: 'no-preference', prefersReducedMotion: false },
+    { query: 'reduce', prefersReducedMotion: true },
+  ])(
+    'should return a boolean indicating whether or not the user prefers reduced motion',
+    ({ query, prefersReducedMotion }) => {
+      window.matchMedia = jest.fn().mockImplementation((q) => ({
+        matches: q === `(prefers-reduced-motion: ${query})`,
+        media: '',
+        onchange: jest.fn(),
+        addEventListener: jest.fn(),
+        removeEventListener: jest.fn(),
+      }));
+
+      const { result } = renderHook(() => usePrefersReducedMotion());
+      expect(result.current).toBe(prefersReducedMotion);
+    }
+  );
+
+  it.each([
+    { query: 'no-preference', prefersReducedMotion: false },
+    { query: 'reduce', prefersReducedMotion: true },
+  ])(
+    'should set up listeners for changes to the reduced motion preference',
+    ({ query, prefersReducedMotion }) => {
+      const addEventListenerMock = jest.fn();
+      const removeEventListenerMock = jest.fn();
+
+      window.matchMedia = jest.fn().mockImplementation((q) => ({
+        matches: q === `(prefers-reduced-motion: ${query})`,
+        media: '',
+        onchange: jest.fn(),
+        addEventListener: addEventListenerMock,
+        removeEventListener: removeEventListenerMock,
+      }));
+
+      const { result, unmount } = renderHook(() => usePrefersReducedMotion());
+      expect(result.current).toBe(prefersReducedMotion);
+
+      expect(addEventListenerMock).toHaveBeenCalledTimes(1);
+
+      unmount();
+
+      expect(addEventListenerMock).toHaveBeenCalledTimes(1);
+      expect(removeEventListenerMock).toHaveBeenCalledTimes(1);
+    }
+  );
+});
--- a/__tests__/hooks/usePrefersReducedMotion.spec.ts
+++ b/__tests__/hooks/usePrefersReducedMotion.spec.ts
@@ -3,6 +3,15 @@ import { renderHook } from '@testing-library/react-hooks';
 import { usePrefersReducedMotion } from '../../src/hooks/usePrefersReducedMotion';
 
 describe('usePrefersReducedMotion', () => {
+  afterAll(() => {
+    // jsdom does not implement window.matchMedia, so we mock it in this suite.
+    // Ensure this is reset to undefined at the end of the suite.
+    // TS will complain about the assignment to undefined, but we can safely ignore it here.
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    window.matchMedia = undefined;
+  });
+
   it.each([
     { query: 'no-preference', prefersReducedMotion: false },
     { query: 'reduce', prefersReducedMotion: true },
--- a/src/hooks/shared.ts
+++ b/src/hooks/shared.ts
@@ -96,7 +96,50 @@ export const onComplete = <E extends HTMLElement | SVGElement>({
   onAnimationComplete?.();
 };
 
-interface CheckCacheParams {
+interface DeriveAccessibleFromToParams {
+  prefersReducedMotion: boolean;
+  from: CSSProperties;
+  to: CSSProperties;
+  reducedMotion?: {
+    from: CSSProperties;
+    to: CSSProperties;
+  };
+}
+
+/**
+ * The deriveAccessibleFromTo function determines the correct from / to
+ * combination to use for accessible animations. The algorithm:
+ *
+ * - If a user does not prefer reduced motion, return the base from / to combination.
+ * - If the user does prefer reduced motion and there's a reducedMotion config applied,
+ *   return the reducedMotion config.
+ * - Else, immediately animate to the "to" state on the first frame to avoid animations
+ *   when a user explicitly prefers reduced motion.
+ */
+export const deriveAccessibleFromTo = ({
+  prefersReducedMotion,
+  from,
+  to,
+  reducedMotion,
+}: DeriveAccessibleFromToParams): CSSPairs => {
+  if (!prefersReducedMotion) {
+    return {
+      from,
+      to,
+    };
+  }
+
+  if (prefersReducedMotion && reducedMotion) {
+    return reducedMotion;
+  }
+
+  return {
+    from: to,
+    to,
+  };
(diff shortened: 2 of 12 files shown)


src/hooks/usePrefersReducedMotion.ts
```diff
@@ -0,0 +1,30 @@
+import { useState, useEffect } from 'react';
+
+const query = '(prefers-reduced-motion: no-preference)';
+
+const getInitialState = () =>
+  typeof window === 'undefined'
+    ? true
+    : !window.matchMedia('(prefers-reduced-motion: no-preference)').matches;
```
upatel32: you could use `query` here?

src/hooks/shared.ts
```diff
@@ -95,3 +95,37 @@ export const onComplete = <E extends HTMLElement | SVGElement>({
...
+  cache: MutableRefObject<AnimationCache>;
+  index: number;
+  from: CSSProperties;
+  to: CSSProperties;
+}
+
+/**
+ * The checkAnimationCache functions inspects the hook's local animation cache to see
+ * if there are animation properties applied. It also checks if the cached properties
+ * match that of the to configuration, signaling it's safe to animate. Otherwise, use
```
upatel32: nit: typo
`match that of the to configuration`

upatel32 (approved):

src/hooks/shared.ts
```diff
@@ -95,3 +95,37 @@ export const onComplete = <E extends HTMLElement | SVGElement>({
...
+ */
+export const checkAnimationCache = ({
+  cache,
+  index,
+  from,
+  to,
+}: CheckCacheParams): CSSPairs => {
+  const cachedFrom = cache.current.get(index);
+
+  if (cachedFrom && Object.keys(cachedFrom) === Object.keys(to)) {
```
gingerrific: I'm not 100% if this is the intent or not (or if I'm missing something else), but won't this always be `false` due to `Object.keys` returning an array?
```
const foo = { bar: 1, baz: 2 }
Object.keys(foo) === Object.keys(foo) // false
```

parkerziegler: 🤦 Great catch @gingerrific! Yep, the intent is basically to see "Do we have cached animation values for every property specified in `to`. We do? Great, let's use those cached values. We don't? Ah, just use `from` and `to`." Will add a more stable check here!

src/hooks/shared.ts
parkerziegler: This is intentional language, but I see how its confusing. I'll put `to` in quotes to match it clear (it's referencing the `to` property specified in a `renature` hook).

src/hooks/usePrefersReducedMotion.ts
parkerziegler: Good catch!

gingerrific:
@parkerziegler I think my main concern/question here is around the default behavior and required configuration. You mention 
> we go for accessible animations as the default behavior  

which is part true, but it requires a fully defined `reducedMotion` prop object to work. I think the idea of allowing the implementer to define what they want in the case of their user having the `reduced motion` OS setting toggled is a very much a dev friendly approach, but shouldn't we default to *no animations* in the case of the user not providing a `reducedMotion` value to truly default to the above mentioned accessibility?

Something roughly like:
```
const getTransitionProps = (
  reducedMotionProps: {
    from: CSSProperties,
    to: CSSProperties,
  } | undefined,
  baseFrom: CSSProperties,
  baseTo: CSSProperties,
  prefersReducedMotion: boolean
) => {
  // early exit if the toggle is disabled
  if (!prefersReducedMotion) {
    return {
      from: baseFrom,
      to: baseTo
    }
  }

  if (reducedMotionProps && Object.keys(reducedMotionProps).length > 0) {
    // if implementation includes a config object for reducedMotion from|to, use it here
    // obviously this is bare minumum of "validation" of a prop object, but for illustrative purposes
    return reducedMotionProps
  } else {
    // if no reducedMotion object is provided, but the prefersReducedMotion OS toggle is clicked, use the provided "to" value to skip animation
    return {
      from: baseTo,
      to: baseTo
    }    
  }
}
```
and then switching this in for the lines [57 - 61](https://github.com/FormidableLabs/renature/pull/132/files?file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-f156797a46433d6aca8ec0ec7eda9fc277ced3d32eaba258a1fddbb539f0fc57R57-R61) in `useForceGroup.ts` ~
```
const { from, to } = getTransitionProps(props.reducedMotion, props.from, props.to, prefersReducedMotion)
```

Otherwise, as you can see in the gif in your description, the component still "animates" when you click the OS preference, it's just opacity popping in instead of a spin

__tests__/hooks/usePrefersReducedMotion.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { renderHook } from '@testing-library/react-hooks';
+
+import { usePrefersReducedMotion } from '../../src/hooks/usePrefersReducedMotion';
+
```
gingerrific, line 4: I don't necessarily have a great alternative or reason they should be removed, but these do feel a little bit like they're just testing browser and hooks functionality instead of our code 🤷

src/hooks/shared.ts
upatel32: OH right I see what you mean

src/hooks/shared.ts
parkerziegler: Resolved in latest commit!

__tests__/hooks/usePrefersReducedMotion.spec.ts
parkerziegler: I think they're still meaningful! The first test in the suite validates that the hook returns the _correct_ `boolean` for the given scenarios we're testing, i.e. that encountering `(prefers-reduced-motion: reduce)` results in `true` being returned to the caller and that `(prefers-reduced-motion: no-preference)` results in `false` (and not vice versa).

I agree the second test is more "Ok, `useEffect` can probably handle this just fine." But still good to verify the event listeners get attached and cleaned up on unmount and that we don't risk leaving a rogue event listener hanging around causing a memory leak.

__tests__/hooks/usePrefersReducedMotion.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+describe('usePrefersReducedMotion', () => {
+  it.each([
+    { query: 'no-preference', prefersReducedMotion: false },
+    { query: 'reduce', prefersReducedMotion: true },
+  ])(
+    'should return a boolean indicating whether or not the user prefers reduced motion',
+    ({ query, prefersReducedMotion }) => {
+      window.matchMedia = jest.fn().mockImplementation((q) => ({
```
gingerrific, line 21: did you run into the same issue where `jsdom` doesn't actually emulate this property yet so you can't do a proper `jest.spy` or similar style of mock implementation? not sure if its worth calling out the limitation. 

Also should we have an `afterAll` that resets this to `window.matchMedia = undefined` or anything to avoid possible pollution/bugs elsewhere. The potential feels low, but seems like it'd be good to watch our butts so to speak

__tests__/hooks/usePrefersReducedMotion.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+
+      window.matchMedia = jest.fn().mockImplementation((q) => ({
+        matches: q === `(prefers-reduced-motion: ${query})`,
+        media: '',
+        onchange: jest.fn(),
+        addEventListener: addEventListenerMock,
+        removeEventListener: removeEventListenerMock,
+      }));
+
+      const { result, unmount } = renderHook(() => usePrefersReducedMotion());
```
gingerrific, line 51: all the other test things said, this hooks library seems rad. Nice work!

gingerrific (approved):

__tests__/hooks/usePrefersReducedMotion.spec.ts
parkerziegler: Yep, exactly. Jest has a good guide on `window.matchMedia` specifically, and this is adapted from that: https://jestjs.io/docs/manual-mocks#mocking-methods-which-are-not-implemented-in-jsdom

Good shout on the `afterAll`, I'll add that in.

__tests__/hooks/usePrefersReducedMotion.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+
+import { usePrefersReducedMotion } from '../../src/hooks/usePrefersReducedMotion';
+
+describe('usePrefersReducedMotion', () => {
+  afterAll(() => {
+    // jsdom does not implement window.matchMedia, so we mock it in this suite.
+    // Ensure this is reset to undefined at the end of the suite.
+    // TS will complain about the assignment to undefined, but we can safely ignore it here.
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
gingerrific, line 11: nit: prefer `// @ts-expect-error`

PR #136, opened 2021-04-02
feat: pass physics values to onFrame callback

parkerziegler:
This PR fulfills a feature request from @littlemilkstudio.

The `onFrame` callback now receives the `position`, `velocity`, and `acceleration` values in addition to the `progress` value. For `useGravity2D`, which never received a `progress` value because the orbit animation is infinite by nature, `onFrame` also receives these vectors.

This should make accessing these physics values for us in Three.js / r3f fairly trivial. The API would look something like the following for `useGravity2D`:

```typescript
useGravity2D({
  config: {
    // Physics configuration values
  },
  onFrame: ({ position, velocity, acceleration }) => {
    // Execute some code with the raw physics values, all of which are tuples representing vectors.
  }
});
```

For the other hooks, access will be slightly different. `progress` is the first argument passed to `onFrame`. I didn't want to group `progress` with the physics values to prevent a breaking API change.

```typescript
useFriction({
  config: {
    // Physics configuration values
  },
  onFrame: (progress, { position, velocity, acceleration }) => {
    // Execute some code with the raw physics values, all of which are two tuples representing vectors.
  }
});
```

Hopefully this can be helpful for our Three.js shot! Let me know if there are some missing additional values that would be helpful here.

### DIFF
--- a/__tests__/rAF/update.spec.ts
+++ b/__tests__/rAF/update.spec.ts
@@ -102,6 +102,7 @@ describe('update', () => {
     const mockMoverState: Partial<Entity> = {
       velocity: [2, 0],
       position: [5, 0],
+      acceleration: [-1.5, -2],
     };
 
     const element: StatefulAnimatingElement<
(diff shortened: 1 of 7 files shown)


parkerziegler:
FYI @littlemilkstudio I know you're on the road so merging this as is!

maxyinger:
Looks great! thanks!

PR #138, opened 2021-04-15
docs: improve table UI for API page with custom MDX components

parkerziegler:
This PR improves the rendering of our table UI on the API page.

Currently, our tables are incredibly hard to read due to the `table` layout we use.

**Desktop**

image

**Mobile (Emulated Pixel 2)**

image

A more suitable layout for this type of information (API documentation) would be in a list format a la [`framer-motion`](https://www.framer.com/api/motion/) or [Stripe](https://stripe.com/docs/api). To that end, this PR introduces a new list UI for Markdown tables. This is made possible by the fact that we can intercept the rendering of a Markdown table using MDX to instead render an unordered list.

**Desktop**

image

**Mobile (Emulated Pixel 2)**

image

In addition, I made a few changes to our language in documentation to remove references to Reason since we now use ReScript as our second source language.


gingerrific (approved):
Overall looks good. 

Only nits would be in `api.md`. 
1. For [`CSSPairs`](https://github.com/FormidableLabs/renature/pull/138/files#diff-15108efcf187ae89e032690449645f7df8d753c4cfd791dfd814b2d85453641eL20) it'd be good to have this either link to your definition, or since it isn't used anywhere else, could just be:
```
{
  from: CSSProperties;
  to: CSSProperties;
}
```
2. Similarly, I think it'd be great to have a call out for what [`MotionVector`](https://github.com/FormidableLabs/renature/pull/138/files#diff-15108efcf187ae89e032690449645f7df8d753c4cfd791dfd814b2d85453641eL25) type is - especially as 
```
interface MotionVectors {
  position: Vector<number>;
  velocity: Vector<number>;
  acceleration: Vector<number>;
}
```
isn't particularly intuitive

parkerziegler:
@gingerrific agreed! The one design challenge there is multi-line code blocks inside of Markdown tables are a bit of a 🐻 The best solution I found for representing them is inlining `pre` tags with `<br />` elements for full line breaks and `&nbsp;` for indentation. Even then, the UI suffers from aligning a four or five line code snippet with a single property name.

I do like the idea of linking to the actual `interface` definition in the source, so that you can click on it and immediately get taken to the type definition.

`framer-motion` seems to solve this all by having a right-aligned code example for each property. I think that could be a reasonable longer term strategy, but would likely mean abandoning Markdown tables entirely for storing this information since making multi-line code samples is such a pain. Ideally that can happen if the docs sites ever truly adopt MDX over our current setup (strict MD pages which get run through a post-processor to allow for some MDX capabilities).

PR #153, opened 2021-05-29
test: add example unit tests for useFriction

parkerziegler:
This PR adds example unit tests for the `useFriction` hook. The goal here is to demonstrate some of the strategy around unit testing this hook so that other contributors can write similar tests to help up coverage and reliability of `renature`.

At first, I planned to use [`@testing-library/react-hooks`](https://github.com/testing-library/react-hooks-testing-library) for these tests, but `renature`'s design introduced a few challenges:

1. A hook like `useFriction` returns a `ref`, which gets attached to a DOM node. Unless we have a DOM node to attach the `ref` to, its value will be `null`. This means we need more structure than `@testing-library/react-hooks` provides — we need to both call the hook _and_ attach it to a DOM node so we can assert on it.
2. We are primarily interested in testing how animating nodes behave from an end user's perspective rather than testing the behavior of the hook itself. For this reason, I opted to use `@testing-library/react` so we can assert on the animating element itself (i.e. particulars of its `style` attribute).

These tests did help me uncover one undesirable edge case with how animations come to an end. I'll write up an issue to track that as well.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -8,6 +8,7 @@
         }
       }
     ],
+    "@babel/preset-react",
     "@babel/preset-typescript"
   ]
 }
--- a/__tests__/hooks/useFriction.spec.tsx
+++ b/__tests__/hooks/useFriction.spec.tsx
@@ -1,0 +1,176 @@
+import React from 'react';
+import { render, waitFor } from '@testing-library/react';
+
+import { useFriction } from '../../src';
+
+/**
+ * Wait an arbitrary amount of time for the animation to run.
+ * We aren't concerned about what particular state the animation
+ * has reached, just that some animation is actively happening.
+ *
+ * @returns – a Promise that resolves to true after 500ms.
+ */
+const waitForAnimation = async (
+  { timeout, waitForTimeout } = { timeout: 500, waitForTimeout: 1000 }
+): Promise<boolean> =>
+  await waitFor(
+    () =>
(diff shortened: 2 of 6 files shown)


.babelrc
```diff
@@ -8,6 +8,7 @@
         }
       }
     ],
+    "@babel/preset-react",
```
parkerziegler, line 11: Required to enable tests that use JSX, otherwise we'll get complaints from Jest about unrecognized characters (like `<`).

__tests__/hooks/useFriction.spec.tsx
```diff
@@ -0,0 +1,176 @@
...
+    // meaning that the 1000 ms delay will be exceeded and the animation should be running.
+    await waitForAnimation({ timeout: 1000, waitForTimeout: 1500 });
+
+    const currentOpacity = parseFloat(node.style.opacity);
+
+    expect(currentOpacity).toBeGreaterThan(0);
+    expect(currentOpacity).toBeLessThan(1);
+  });
+
+  it('should not apply translateZ on an element animating the transform property if disableHardwareAcceleration is true', () => {
```
gingerrific: nit, I might move this test up next to the other `translate` test just for more clarity on why this omission is meaningful 🤷

gingerrific (approved):
lgtm

